PR #6, opened 2021-09-05
Setup ci on github actions

laurentS:
This PR sets up a simple CI workflow that:
- installs dependencies with poetry
- runs black and mypy
- runs pytest tests
- does all the above in all supported python versions

This PR should be merged only after https://gitlab.com/meltano/sdk/-/merge_requests/150 has been merged (mypy won't succeed without it, but I've verified locally that it does work if the `py.typed` file is manually added to the dependency)

Once we have a decent CI setup here, it would be useful to copy the workflow file into the SDK's template, so that any new tap can use the same logic.

### DIFF
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -1,0 +1,41 @@
+name: Test tap-github
+
+on:
+  push
+  # uncomment this once the workflow is passing
+  # push:
+  #   branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  tests:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Poetry
+      uses: snok/install-poetry@v1
+      with:
+        # Version of Poetry to use
+        version: 1.1.8
+    - name: Install dependencies
+      run: |
+        poetry install
+    - name: Check formatting with black
+      run: |
+        poetry run black --check .
+    - name: Check typing annotations with mypy
+      run: |
+        poetry run mypy .
+    - name: Test with pytest
+      run: |
+        poetry run pytest
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -29,7 +29,7 @@ jobs:
         poetry run black --check .
     - name: Check typing annotations with mypy
       run: |
-        poetry run mypy .
+        poetry run mypy . --ignore-missing-imports
     - name: Test with pytest
       run: |
         poetry run pytest
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -1,9 +1,6 @@
 name: Test tap-github
 
-on:
-  - push
-  - pull_request:
-      branches: [ main ]
+on: [push, pull_request]
 
 jobs:
   tests:
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -1,12 +1,9 @@
 name: Test tap-github
 
 on:
-  push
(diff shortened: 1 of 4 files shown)


tap_github/client.py
```diff
@@ -26,7 +26,7 @@ def http_headers(self) -> dict:
         """Return the http headers needed."""
         headers = {"Accept": "application/vnd.github.v3+json"}
         if "user_agent" in self.config:
-            headers["User-Agent"] = self.config.get("user_agent")
+            headers["User-Agent"] = self.config.get("user_agent")  # type: ignore
```
laurentS, line 29: Is there a way to make mypy aware of the jsonschema for the config?

laurentS:
The checks are [passing](https://github.com/laurentS/tap-github-1/actions/runs/1206263946) on my fork.

tap_github/client.py
aaronsteers: There may be a json schema manner that is better, but this at least is explicit. Unlike in other languages, ["cast()"](https://mypy.readthedocs.io/en/stable/casts.html) is purely a type hint - but it nonetheless makes it easy for the reader to understand the expected behavior.

```suggestion
            headers["User-Agent"] = cast(str, self.config.get("user_agent"))
```

pyproject.toml
```diff
@@ -15,9 +15,10 @@ singer-sdk = "^0.3.4"
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.1.2"
-mypy = "^0.812"
+mypy = "^0.910"
 flake8 = "^3.9.2"
 black = {version = "^21.5b1", allow-prereleases = true}
+types-requests = "^2.25.6"
```
aaronsteers, line 21: @laurentS - Should we install this type stub package also in the sdk core package?

cc @edgarrmondragon

.github/workflows/test_tap.yml
```diff
@@ -0,0 +1,35 @@
+name: Test tap-github
+
+on: [push, pull_request]
+
+jobs:
+  tests:
```
aaronsteers, line 6: @laurentS - This is a very elegant and simple CI flow. Nicely done. 👍 
Any objection to us replicating this upstream and adapting into the SDK cookiecutter template? 

cc @edgarrmondragon

laurentS: By all means, please do! I mentioned that same idea in my initial comment, it'd be great to reuse it as much as possible :)

pyproject.toml
laurentS: if the template relies on `requests`, then yes, as `mypy` will be looking for the type declarations as well.

tap_github/client.py
laurentS: It does look better indeed. Committed below. Thanks for the suggestion!

.github/workflows/test_tap.yml
```diff
@@ -0,0 +1,35 @@
...
+        poetry install
+    - name: Check formatting with black
+      run: |
+        poetry run black --check .
+    - name: Check typing annotations with mypy
+      run: |
+        poetry run mypy . --ignore-missing-imports
+    - name: Test with pytest
+      run: |
+        poetry run pytest
```
edgarrmondragon, line 35: @laurentS If the first one fails, do the other checks still run?

laurentS: No, it stops immediately. Github actions will also abort all jobs in the matrix if one of them fails, for instance if python 3.6 fails, all parallel jobs will be cancelled no matter how far they got (unless they completed before). You can see an example of this [here](https://github.com/laurentS/slowapi/runs/3450829365?check_suite_focus=true) (I just reused the code I wrote there, the CI definition is exactly the same).
I don't know if there's a way to change that behaviour. Given that the tests run for about a minute, I haven't bothered digging :)

edgarrmondragon: > if python 3.6 fails, all parallel jobs will be cancelled no matter how far they got

I didn't know that, thanks for sharing!

> the tests run for about a minute

You're right in that. And the order is OK then. First code style, then type annotations, then tests 👍 

cc @aaronsteers

aaronsteers (approved):
Looks great! I'm happy to approve.

@edgarrmondragon - Feel free to merge if there's nothing else outstanding.

Issue #5, opened 2021-09-03
Allow use cases where list of `projects` or `organizations` is already known


aaronsteers:
Currently, this was built around the `searches` config option, which allows Github repo searches to drive dynamic repo selection.

I suggest we make three vectors for selection, each which would be mutually exclusive during setting validation:

1. `searches` - (existing) finds repos based on the search API
2. `repositories` - (hi-pri) takes a list of one or more repositories, such as `['MeltanoLabs/tap-github']`
3. `organizations` - (low-pri) takes a list of one or more organizations, such as `['MeltanoLabs']`

laurentS:
I got started on `repositories` this afternoon (EU time).

aaronsteers:
@laurentS - Fantastic! Thanks!

aaronsteers:
@laurentS - Thanks for your work adding `repositories` as a top-level config setting in #7. For the organizations use case, do you think you could add one of the sample queries you shared with me into the Readme? It occured to me that we could solve this use case by simply making it straightforward on how to query for everything under a specific org - and relying on the existing search-oriented features for that use case.

What do you think?

aaronsteers:
Closing this as resolved. Repositories is now supported as a config option, and `organizations` can be supported via a search using Github query syntax to include only the requested org(s).

PR #7, opened 2021-09-06
Add basic support for a list of repos as input

laurentS:
@aaronsteers  This is a draft PR for feedback, as I'm not familiar enough with the SDK to make optimal use of it :)

I've added basic support for passing a list of repos to fetch data for instead of passing a search query. It only works for the `repositories` stream now, and I sort of abuse the `search/repositories` API endpoint by using its `repo:xxx` query mode, which returns exactly the repo passed by name ([docs](https://docs.github.com/en/github/searching-for-information-on-github/searching-on-github/searching-for-repositories#search-by-repository-name-description-or-contents-of-the-readme-file)).
A few questions:
- I didn't find a way to do an "XOR" sort of declaration in jsonschema for the config (to force having `searches XOR repositories`). There seems to be a `oneOf` in JSON Schema [itself](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.2.1.3), but I don't know if it makes more sense to try and implement that in `typing.py` or just raise an exception for invalid config in `tap-github` itself.
- in `test_tap.py` I try to validate that I'm getting the correct number of data records, but I suspect you might have a helper function for that purpose. I'm not sure what the best pattern for testing the output of a tap is.
- The test I added actually fetches records from the github API. While I think it provide some level of safety against possible future changes in the API (because the tests would break in such a case), it also makes the tests brittle because of rate limiting (I got bitten!) or whatever networking problem. It might make more sense to use mock API responses for faster tests and less variability though.
- the response content is slightly different between the `repos/{org}/{repo}` and `search/repositories` endpoints, with the former adding a few more fields, mainly about the organisation holding the repo. I don't know if it makes sense to have 2 different schemas here, based on what the config params were?

Happy to hear your thoughts on these points before I add support for other streams.


laurentS:
Related to issue #5

tap_github/streams.py
```diff
@@ -22,16 +22,27 @@ def get_url_params(
         """Return a dictionary of values to be used in URL parameterization."""
         assert context is not None, f"Context cannot be empty for '{self.name}' stream."
         params = super().get_url_params(context, next_page_token)
-        params["q"] = context["search_query"]
+        if "search_query" in context:
+            # we're in search mode
+            params["q"] = context["search_query"]
+        else:
+            # we are in repository list mode
+            params["q"] = f"repo:{context['org']}/{context['repo']}"
```
aaronsteers, line 30: For this use case, I think there's a more performant `repositories` endpoint to use, rather than the `search` endpoint.

aaronsteers (changes requested):
@laurentS - Your question in #8 was apt but I think moreso for the repositories themselves. 
Can we use [this endpoint](https://docs.github.com/en/rest/reference/repos#get-a-repository) when receiving a specific list of repos in place of search?

tap_github/streams.py
aaronsteers: In line 59, where we have `path = "/search/repositories"`, I think something like this might do the trick:

```py
    @property
    def path(cls) -> str:
        if "searches" in self.config:
            return "/search/repositories"
        else:
            # the `repo` and `org` args will be parsed from the partition's `context`
            return "/repos/{org}/{repo}"         
```

aaronsteers: > According to the docs, the ?q=repo:org/repo should return exactly the repo you specify (I tried searching for one that does not exist, the call throws an error). A bit of testing seems to confirm that.

Awesome. Given this ☝️, I'm actually totally fine with either approach. While the `repos` endpoint seems more purpose built for one repo at a time, there are pros and cons of each approach which we could always work through in a future update.

aaronsteers (approved):

aaronsteers:
@laurentS - Do you mind updating from main now that #6 is merged? Would love to see the CI tests before merging.

laurentS:
it's not perfect, but the "tests" are passing. I has to tweak them a bit when checking the number of records produced, because the stream selection wasn't working (as discussed on slack), despite upgrading to the latest sdk you released yesterday. The tap's output looks ok though. What do you suggest as a method for testing a bit more than just "it runs without crashing"?

aaronsteers (approved):
@laurentS - This is looking good to me. Approving. I think we can improve incrementally from here, especially on the testing side, which you've called out.

Perhaps in a new issue (or issues) we can track next steps?

laurentS:
Sounds good to me. Thanks for merging so quickly, much appreciated!

Issue #8, opened 2021-09-06
Using different API endpoints for a "same" stream


laurentS:
Currently `issue_comments` are fetched from [`/repos/{owner}/{repo}/issues/{issue_number}/comments`](https://docs.github.com/en/rest/reference/issues#get-an-issue-comment) as a child stream of `issues`. This allows fetching only comments for a single issue.

Our use case is more suited to fetching the same data at the repo level, using the [`/repos/{owner}/{repo}/issues/comments`](https://docs.github.com/en/rest/reference/issues#list-issue-comments-for-a-repository) API endpoint. It returns the same data for each comment, according to the docs.

The main value for us in using the second solution is that we get more data for the same API usage quota: the typical issue will have far fewer than 100 comments on it, so we get comments for multiple issues in one call. We do not run this as a child stream, instead we feed it a list of repos for which we want data (as in #5), and sync only this stream. This allows us better control over API usage limits.

Is there a way to have both streams coexist in the tap so that either can be used if needed, but the default setup would not double fetch data? (what I mean is that `issue_comments_by_repo` could be a child stream of `repositories`, and `issue_comments_by_issue` a child of `issues`, itself a child of `repositories`, so if fetching the whole catalog, we would fetch some data twice).

I think there are other streams which have a similar situation, so this is probably a slightly more general question than just this stream (in a similar fashion, `repositories` can be obtained through the `search` endpoint, or the `repos` endpoint, although the data returned is slightly different). Another example is `issue_events` which can be fetch by issue or by repo.

aaronsteers:
@laurentS - I wonder if it's best to just support `issue_comments` in this case as a child of `repo` (driven by the `{repo}/issues/coments` endpoing _instead of_ being a child of `issues`. Performance would be improved, I think, to make a paginated repo-level call, rather than a separate round of calls for each issue, and it would also speak to your points 

 At the time when I was developing the first version of this tap, we were needing an example case to prove parent-child and grandparent-parent-child relationships for the SDK, and for this reason we probably chose a less-than-ideal retrieval path - in the interest of proving the more complex use case.

Given all the above, I think I'd be inclined to take your proposed logic in place of how we have it set now. Wdyt?

Ry-DS:
Was this resolved in #9? We could close then if so :eyes:

laurentS:
Indeed, thanks for pointing it out.

PR #9, opened 2021-09-10
Add issues and issue_comments streams

laurentS:
This PR does a couple of things:
- update the `issues` stream schema to match the docs (and actual API output) a bit more closely. I've not included *all* fields, namely excluding the "obvious" ones that can be computed trivially by appending a common postfix to the base url.
- modify the `issue_comments` stream to be a child of `repositories` as discussed in #8 and update the schema like above

Depending on how quickly this gets merged, I might add `issue_events` here, or in a separate PR, following the same logic.

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -280,9 +280,14 @@ class IssueCommentsStream(GitHubStream):
                 params["since"] = since
         return params
 
+    def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
+        row['issue_number'] = int(row["issue_url"].split('/')[-1])
+        return row
+
     schema = th.PropertiesList(
         th.Property("id", th.IntegerType),
         th.Property("node_id", th.StringType),
+        th.Property("issue_number", th.IntegerType),
         th.Property("repo", th.StringType),
         th.Property("org", th.StringType),
         th.Property("issue_url", th.IntegerType),
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -130,17 +130,116 @@ class IssuesStream(GitHubStream):
 
     schema = th.PropertiesList(
         th.Property("id", th.IntegerType),
+        th.Property("node_id", th.StringType),
+        th.Property("url", th.StringType),
+        th.Property("html_url", th.StringType),
         th.Property("repo", th.StringType),
         th.Property("org", th.StringType),
-        th.Property("issue_number", th.IntegerType),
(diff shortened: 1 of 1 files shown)


tap_github/streams.py
```diff
@@ -179,11 +282,25 @@ def get_url_params(
 
     schema = th.PropertiesList(
         th.Property("id", th.IntegerType),
+        th.Property("node_id", th.StringType),
         th.Property("repo", th.StringType),
         th.Property("org", th.StringType),
-        th.Property("issue_number", th.IntegerType),
+        th.Property("issue_url", th.IntegerType),
```
aaronsteers, line 288: I think the `issue_number` property was previously coming from parent context. What do you think of adding back via `post_process()` method? 

Something like `"/".split(record["issue_url"])[-1]` might do the trick.

tap_github/streams.py
```diff
@@ -130,28 +130,131 @@ def http_headers(self) -> dict:
...
+        th.Property("url", th.StringType),
+        th.Property("html_url", th.StringType),
         th.Property("repo", th.StringType),
         th.Property("org", th.StringType),
-        th.Property("issue_number", th.IntegerType),
+        th.Property("number", th.IntegerType),
         th.Property("updated_at", th.DateTimeType),
         th.Property("created_at", th.DateTimeType),
-        # th.Property("closed_at", th.DateTimeType),  # Nulls causing parse error
+        th.Property("closed_at", th.DateTimeType),
```
aaronsteers, line 141: Thanks for restoring this. I think the null behavior problem came up very early in testing and I forgot to retest and restore the field.

aaronsteers (changes requested):
I think we can merge this quickly. Just one requested change regarding issue number on the comments stream. Let me know what you think. Thanks!

tap_github/streams.py
laurentS: I like that! I'll add it in. It does seem handy to have the issue number available.

laurentS:
I've update the code as per your comments. I've been bothered by #11 but it's not critical for merging this, it just lets too much data through I think.

aaronsteers:
> I've update the code as per your comments. I've been bothered by #11 but it's not critical for merging this, it just lets too much data through I think.

Responded on #11. Yes, I think that's an SDK issue and not needed to be resolved here necessarily. Thanks for raising.

aaronsteers (approved):
Looks great! 👍

PR #12, opened 2021-09-11
Fix some errors in stream schema definitions

laurentS:
I let some errors slip through in #9 sorry. These definitions actually seem to work when running the tap against the API (I'll blame it on hasty copy/pasting :sweat_smile: )


aaronsteers (approved):

aaronsteers:
Looks good! Thanks!

PR #13, opened 2021-09-12
Add commits stream

laurentS:
This PR adds another stream to the tap, for commits.
I improvised a bit for the replication key, as the API does not provide a top-level timestamp field, and couldn't figure out a way to use a nested one.

There is no testing either, but I've run the tap against `target-postgres` and the data appears as expected, which is better than nothing :)

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -375,3 +375,113 @@ class IssueCommentsStream(GitHubStream):
             ),
         ),
     ).to_dict()
+
+
+class CommitsStream(GitHubStream):
+    """
+    Defines the 'Commits' stream.
+    The stream is fetched per repository to maximize optimize for API quota
+    usage.
+    """
+
+    name = "commits"
+    path = "/repos/{org}/{repo}/commits"
+    primary_keys = ["node_id"]
+    replication_key = "commit_timestamp"
+    parent_stream_type = RepositoryStream
+    state_partitioning_keys = ["repo", "org"]
+    ignore_parent_replication_key = True
+
+    def get_url_params(
+        self, context: Optional[dict], next_page_token: Optional[Any]
+    ) -> Dict[str, Any]:
+        """Return a dictionary of values to be used in URL parameterization."""
+        params = super().get_url_params(context, next_page_token)
+        if self.replication_key:
+            since = self.get_starting_timestamp(context)
+            if since:
+                params["since"] = since
+        return params
+
+    def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
+        """
+        Add a timestamp top-level field to be used as state replication key.
+        It's not clear from github's API docs which time (author or committer)
+        is used to compare to the `since` argument that the endpoint supports.
+        """
+        row["commit_timestamp"] = row["commit"]["committer"]["date"]
+        return row
+
+    schema = th.PropertiesList(
+        th.Property("node_id", th.IntegerType),
+        th.Property("url", th.StringType),
+        th.Property("sha", th.StringType),
+        th.Property("html_url", th.StringType),
+        th.Property("commit_timestamp", th.DateTimeType),
+        th.Property(
+            "commit",
+            th.ObjectType(
+                th.Property(
+                    "author",
+                    th.ObjectType(
+                        th.Property("name", th.StringType),
+                        th.Property("email", th.StringType),
+                        th.Property("date", th.DateTimeType),
+                    ),
+                ),
+                th.Property(
+                    "committer",
+                    th.ObjectType(
+                        th.Property("name", th.StringType),
+                        th.Property("email", th.StringType),
+                        th.Property("date", th.DateTimeType),
+                    ),
+                ),
+                th.Property("message", th.StringType),
+                th.Property(
+                    "tree",
+                    th.ObjectType(
+                        th.Property("url", th.StringType),
+                        th.Property("sha", th.StringType),
+                    ),
+                ),
+                th.Property("comment_count", th.IntegerType),
+                th.Property(
+                    "verification",
+                    th.ObjectType(
+                        th.Property("verified", th.BooleanType),
+                        th.Property("reason", th.StringType),
+                        th.Property("signature", th.StringType),
+                        th.Property("payload", th.StringType),
+                    ),
+                ),
+            ),
+        ),
+        th.Property(
+            "author",
+            th.ObjectType(
+                th.Property("login", th.StringType),
+                th.Property("id", th.IntegerType),
+                th.Property("node_id", th.StringType),
+                th.Property("avatar_url", th.StringType),
+                th.Property("gravatar_id", th.StringType),
+                th.Property("html_url", th.StringType),
+                th.Property("type", th.StringType),
+                th.Property("site_admin", th.BooleanType),
+            ),
+        ),
+        th.Property(
+            "committer",
+            th.ObjectType(
+                th.Property("login", th.StringType),
+                th.Property("id", th.IntegerType),
+                th.Property("node_id", th.StringType),
+                th.Property("avatar_url", th.StringType),
+                th.Property("gravatar_id", th.StringType),
+                th.Property("html_url", th.StringType),
+                th.Property("type", th.StringType),
+                th.Property("site_admin", th.BooleanType),
+            ),
+        ),
+    ).to_dict()
+
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -460,7 +460,6 @@ class CommitsStream(GitHubStream):
                         th.Property("sha", th.StringType),
                     ),
                 ),
-                th.Property("comment_count", th.IntegerType),
                 th.Property(
                     "verification",
                     th.ObjectType(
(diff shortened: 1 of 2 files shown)


tap_github/streams.py
```diff
@@ -375,3 +375,112 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+    usage.
+    """
+
+    name = "commits"
+    path = "/repos/{org}/{repo}/commits"
+    primary_keys = ["node_id"]
+    replication_key = "commit_timestamp"
+    parent_stream_type = RepositoryStream
+    state_partitioning_keys = ["repo", "org"]
+    ignore_parent_replication_key = True
```
aaronsteers, line 393: Setting this to `true` means we don't expect the parent (repos, in this case) to have their update timestamp reflective of these additions. Is that correct?

To be fair, the `True` setting is the safest way, since we know we won't miss any updates, but I thought github included this in the repository's last updated timestamp. Maybe my misunderstanding is that only master/main-branch updates get reflected there?

tap_github/streams.py
```diff
@@ -375,3 +375,112 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+                params["since"] = since
+        return params
+
+    def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
+        """
+        Add a timestamp top-level field to be used as state replication key.
+        It's not clear from github's API docs which time (author or committer)
+        is used to compare to the `since` argument that the endpoint supports.
+        """
+        row["commit_timestamp"] = row["commit"]["committer"]["date"]
```
aaronsteers, line 412: 👍 Nice.

tap_github/streams.py
```diff
@@ -375,3 +375,112 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+class CommitsStream(GitHubStream):
+    """
+    Defines the 'Commits' stream.
+    The stream is fetched per repository to maximize optimize for API quota
+    usage.
+    """
+
+    name = "commits"
+    path = "/repos/{org}/{repo}/commits"
+    primary_keys = ["node_id"]
```
aaronsteers, line 389: I haven't referenced the API docs yet but I would expect this to be "commit_ref" or "sha". "Node" or "Node ID" doesn't seem very intuitive and it looks like perhaps those terms are overloaded several times in the stream properties - what do you think?

Is "node" self-explanatory enough in your opinion or should we alias this or perhaps use another property?

tap_github/streams.py
```diff
@@ -375,3 +375,112 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+    def get_url_params(
+        self, context: Optional[dict], next_page_token: Optional[Any]
+    ) -> Dict[str, Any]:
+        """Return a dictionary of values to be used in URL parameterization."""
+        params = super().get_url_params(context, next_page_token)
+        if self.replication_key:
+            since = self.get_starting_timestamp(context)
+            if since:
+                params["since"] = since
+        return params
```
aaronsteers, line 404: I'm not confident in this block. Couple of things...

First, I remember recently we found a bug where `get_starting_timestamp()` was sliding along with state updates. TODO on me (@aaronsteers) to check on that and report back on how this would be affected.

Second, this looks very similar and perhaps redundant with what is already happening in the super. Can we remove?

https://github.com/MeltanoLabs/tap-github/blob/main/tap_github/client.py#L59-L72

tap_github/streams.py
```diff
@@ -375,3 +375,112 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+                ),
+                th.Property("message", th.StringType),
+                th.Property(
+                    "tree",
+                    th.ObjectType(
+                        th.Property("url", th.StringType),
+                        th.Property("sha", th.StringType),
+                    ),
+                ),
+                th.Property("comment_count", th.IntegerType),
```
aaronsteers, line 448: I think we should remove `comment_count` since we want to treat commits as immutable and pull them incrementally based on commit_timestamp. 

```suggestion
```

aaronsteers: A separate stream of commit comments could add back this metric based on comment timestamp.

aaronsteers (changes requested):

tap_github/streams.py
laurentS: The `node_id` is an id that github uses on all of their data elements (users, commits, issues, repos...). It seems to be globally unique, which is why I've used this as the primary key here. I haven't actually found any documentation about this, but looking at a bunch of API call results has led me to think the above.
Theoretically, the SHA is not guaranteed to be unique (see [this SO answer, under note 1](https://stackoverflow.com/a/25128596/1138710)) but I guess the risk of clashes is ridiculously low. So if you think it makes more sense, I don't mind changing here.

tap_github/streams.py
laurentS: Yes, this makes sense! I guess I will just add the `commit_comments` stream as well :)

tap_github/streams.py
laurentS: I think I copy/pasted this from `IssueCommentsStream` as the code is exactly the same. If it is redundant here, I guess we can remove it there as well?

tap_github/streams.py
laurentS: Thanks for clarifying the meaning of this. Just tested this by pushing a single commit to a repo on a throwaway branch:
- `pushed_at` gets updated to the timestamp of the commit
- `updated_at` is not modified.
If pushing to `master/main`, then `updated_at` does get updated as well, but there are a few seconds of gap between these values and the commit time (like this: commit time < pushed_at < updated_at), so on a very active project, I imagine this could lead to records slipping through depending on which timestamp we use.

aaronsteers: Okay, thanks for digging on this.

Based on your explanation, I think you have it correct that we should probably not count on this being authoritative at the parent level. (Keeping as-is to ignore=True, I think, is best.)

tap_github/streams.py
aaronsteers: That's okay. Given your explanation, I think `node_id` makes sense here. I wasn't aware of this design around their API but given that hash collisions do _at least in theory_ happen, I propose keeping `node_id` as you have it.

tap_github/streams.py
aaronsteers: Yes, I think that's right. I don't see anything happening here that isn't happening in the super() method.

tap_github/streams.py
aaronsteers: Found this which was helpful and confirming: https://docs.github.com/en/graphql/guides/using-global-node-ids

aaronsteers:
@laurentS - Do you mind pulling the latest from `main`? Today we released v0.3.8 of the SDK, which resolves the `2to3` issue which was causing failures under the latest version of setuptools.

tap_github/streams.py
```diff
@@ -412,3 +401,101 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
             ),
         ),
     ).to_dict()
+
+
+class CommitsStream(GitHubStream):
```
ericboucher, line 406: @aaronsteers now that we start to have a lot more streams in this file, what is your preferred strategy for ordering them? Alphabetical order? Splitting into files? Eg. one file for issues and issues-related streams?

aaronsteers: @ericboucher - Good question. I'm not sure. I think alpha-sorting makes sense to some degree, except that we would want parent streams (repositories, for instance) so they can easily be referenced by children. One file per stream seems a bit messy but might actually be cleaner - perhaps under a `streams` subfolder? Hard to say. 

What do you think?

tap_github/streams.py
aaronsteers: @laurentS - do you mind removing `commit_comments` from properties here? Since this will be desynced from the source (due to `commits` being immutable and the count not incrementing the replication key `commit_timestamp`), I think we should keep this out of this particular stream and allow users to calculate in another way.

aaronsteers (changes requested):

aaronsteers:
@laurentS - Thanks again for this contribution. Just one change requested and I think this is ready to approve and merge.

tap_github/streams.py
ericboucher: I think one file per "parent" stream would make sense ultimately, potentially with a few shared utils for similar override functions.

ericboucher:
@aaronsteers good catch! This is a bit annoying but I don't think this is critical for us atm.

tap_github/streams.py
aaronsteers: @ericboucher - That works for me. I do also think moving into a streams subfolder would be helpful, and if so, the `streams/__init__.py` file can import all types so the tap could simply import all types from `streams` without having to reference each file.

aaronsteers (approved):
This looks great to me. Thanks for your hard work on this.

tap_github/streams.py
aaronsteers: @ericboucher - Happy to take a re-org as a separate PR if you'd like. I'll go ahead and merge this in the meanwhile. Thanks again.

PR #15, opened 2021-09-15
Add README stream

Ry-DS:
This PR adds a README stream to the tap.
Tested on our own DB and seems to be outputting the correct schema :+1:
image

Concerns:
- Is the `url` a decent primary key here?
- We could consider splitting our our streams to more than one file as things continue to grow

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -161,7 +161,6 @@ class ReadmeStream(GitHubStream):
     name = "readme"
     path = "/repos/{org}/{repo}/readme"
     primary_keys = ["url"]
-    # TODO what would this be? replication_key = "updated_at"
     parent_stream_type = RepositoryStream
     ignore_parent_replication_key = False
     state_partitioning_keys = ["repo", "org"]
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -2,8 +2,6 @@
 
 import requests
 from typing import Any, Dict, Iterable, List, Optional
-
-import requests
 from singer_sdk import typing as th  # JSON Schema typing helpers
 
 from tap_github.client import GitHubStream
@@ -169,8 +167,7 @@ class ReadmeStream(GitHubStream):
     state_partitioning_keys = ["repo", "org"]
 
     def parse_response(self, response: requests.Response) -> Iterable[dict]:
-        json = response.json()
-        yield json
+        return [response.json()]
 
     schema = th.PropertiesList(
         th.Property("type", th.StringType),
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -186,7 +186,7 @@ class ReadmeStream(GitHubStream):
         th.Property("download_url", th.StringType),
         th.Property(
             "_links",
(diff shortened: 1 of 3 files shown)


tap_github/streams.py
```diff
@@ -158,6 +157,40 @@ def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
...
+    primary_keys = ["url"]
+    parent_stream_type = RepositoryStream
+    ignore_parent_replication_key = False
+    state_partitioning_keys = ["repo", "org"]
+
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        return [response.json()]
+
+    schema = th.PropertiesList(
+        th.Property("type", th.StringType),
```
aaronsteers, line 172: Can we add the parent keys `repo` and `org` here? If you declare them in the properties list, they should get populated automatically from parent context.

tap_github/streams.py
```diff
@@ -158,6 +157,40 @@ def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
     ).to_dict()
 
 
+class ReadmeStream(GitHubStream):
+    name = "readme"
+    path = "/repos/{org}/{repo}/readme"
+    primary_keys = ["url"]
```
aaronsteers, line 163: If there should only be one README record result per repo, what about using the primary key of the parent: `["repo", "org"]`

aaronsteers: Do you know (or can you test) how this will behave if the rep has both `README.md` and also something like `README.rst`. This speaks to the primary key question, I think, whether the API will send both records or pick a winner and send just one.

If just one, I think we want the primary key to reflect the repo itself - if two or more is allowed, I think we would key in based on the `url`.

Wdyt?

Ry-DS: I did a quick test, and the api just ends up choosing one over the other and only sends one. It sends this if there is no README: 
```
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-a-repository-readme"
}
```
I'll change the key to the repo/org combination :+1:

aaronsteers:
@Ry-DS - This is ready to go - except that we've got failures in CI due to the twice-upstream issue described here: https://gitlab.com/meltano/sdk/-/issues/213

We are planning to release an updated SDK version today which resolves the compat issue and then we can merge this tomorrow (or later today) once we confirm the tests are 🟢 .

aaronsteers:
> We are planning to release an updated SDK version today which resolves the compat issue and then we can merge this tomorrow (or later today) once we confirm the tests are 🟢 .

We released the SDK 0.3.8 today and I've merged the latest back into the `main` branch here in this repo.

@Ry-DS - Can you merge in the latest from main? Something like `git merge upstream/main`, or similar, followed by a git push should do the trick.

Ry-DS:
> @Ry-DS - Can you merge in the latest from main? Something like `git merge upstream/main`, or similar, followed by a git push should do the trick.

Done :+1:

aaronsteers (approved):

Issue #17, opened 2021-09-17
The "issues" stream is only fetching OPEN issues (and PRs)


ericboucher:
I am interested in puling CLOSED issues and MERGED PRs as well as OPEN ones. The current behavior for the `issues` stream is to pull all OPEN issues (and PRs).

1. I think that by default, the `issues` stream could fetch ALL issues that have been updated since our start date; and we could have an option to only select OPEN issues.

2. Somewhat linked: it could be an opportunity to clearly separate issues and PRs. The issues stream would pull only issues, and a pull_request stream would handle pull_requests.

What do you think?

aaronsteers:
Hi, @ericboucher. Regarding your two points...

> I think that by default, the issues stream could fetch ALL issues that have been updated since our start date; and we could have an option to only select OPEN issues.

100% agreed. I wasn't aware we were missing closed issues/PRs and if true, we should probably treat this as a bug. Happy to review a PR on this if you have any cycles to contribute.

> Somewhat linked: it could be an opportunity to clearly separate issues and PRs. The issues stream would pull only issues, and a pull_request stream would handle pull_requests.

Can you say more about your proposal to have the two distinct streams? My understanding (could be wrong) is that GitHub treats all PRs as issues. Happy to add a stream for PRs if there's a dedicated endpoint for them but I'm not sure it that would lead to removal of them from the Issues sreams.

aaronsteers:
Completed in #21 

🎉

PR #21, opened 2021-09-17
Add PullRequests stream and get all issues and prs regardless of state

ericboucher:
Following up on #17, I would like to propose this solution.

This PR:
- creates a new PullRequests stream that fetches PRs only (regardless of state).
- updates the Issue stream to fetch all issues (and PRs) regardless of state and add a "type" field

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -256,6 +256,7 @@ class IssuesStream(GitHubStream):
         th.Property("comments", th.IntegerType),
         th.Property("author_association", th.StringType),
         th.Property("body", th.StringType),
+        th.Property("type", th.StringType),
         th.Property(
             "user",
             th.ObjectType(
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -237,7 +237,7 @@ class IssuesStream(GitHubStream):
         return headers
 
     def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
-        row["type"] = "pull_request" if row["pull_request"] else "issue"
+        row["type"] = "pull_request" if "pull_request" in row else "issue"
         return row
 
     schema = th.PropertiesList(
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -271,6 +271,23 @@ class IssuesStream(GitHubStream):
             ),
         ),
         th.Property(
+            "reactions",
+            th.ArrayType(
+                th.ObjectType(
+                    th.Property("url", th.StringType),
+                    th.Property("total_count", th.IntegerType),
+                    th.Property("+1", th.IntegerType),
+                    th.Property("-1", th.IntegerType),
+                    th.Property("laugh", th.IntegerType),
+                    th.Property("hooray", th.IntegerType),
+                    th.Property("confused", th.IntegerType),
+                    th.Property("heart", th.IntegerType),
+                    th.Property("rocket", th.IntegerType),
+                    th.Property("eyes", th.IntegerType),
+                ),
+            ),
+        ),
+        th.Property(
             "assignee",
             th.ObjectType(
                 th.Property("login", th.StringType),
@@ -431,20 +448,6 @@ class PullRequestsStream(GitHubStream):
         params["state"] = "all"
         return params
 
-    def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
-        """Return a child context object from the record and optional provided context.
-
-        By default, will return context if provided and otherwise the record dict.
-        Developers may override this behavior to send specific information to child
-        streams for context.
-        """
-        if context is None:
-            raise ValueError("Issue stream should not have blank context.")
-
-        context["issue_number"] = record["number"]
-        context["comments"] = record["comments"]  # If zero, comments can be skipped
-        return context
-
     @property
     def http_headers(self) -> dict:
         """Return the http headers needed.
@@ -509,6 +512,23 @@ class PullRequestsStream(GitHubStream):
                     th.Property("description", th.StringType),
                     th.Property("color", th.StringType),
                     th.Property("default", th.BooleanType),
+                ),
+            ),
+        ),
+        th.Property(
+            "reactions",
+            th.ArrayType(
+                th.ObjectType(
+                    th.Property("url", th.StringType),
+                    th.Property("total_count", th.IntegerType),
+                    th.Property("+1", th.IntegerType),
+                    th.Property("-1", th.IntegerType),
+                    th.Property("laugh", th.IntegerType),
+                    th.Property("hooray", th.IntegerType),
+                    th.Property("confused", th.IntegerType),
+                    th.Property("heart", th.IntegerType),
+                    th.Property("rocket", th.IntegerType),
+                    th.Property("eyes", th.IntegerType),
                 ),
             ),
         ),
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -237,7 +237,7 @@ class IssuesStream(GitHubStream):
         return headers
 
     def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
-        row["type"] = 'pull_request' if row["pull_request"] else 'issue'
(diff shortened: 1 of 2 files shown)


aaronsteers:
@ericboucher - First, thanks very much for this contribution.

A quick first-round review did not surface any red flags. I'm very happy for issues to include "all" states and to be adding the Pull requests as a new stream.

However, I'm not sure on whether it is best to have the issues stream as a superset - including "normal" issues and also pull requests - versus having it be mutually exclusive with pull requests.

Things I considered:

1. field availability: I'm not sure if the issue fields are a strict subset of those available for pulls - or if there are some value-add fields a user may want that are only available in the issues stream. 
2. ontological argument: from github's perspective (at least in the API), PRs are also Issues. 
3. Singer precedent: I did a quick scan of Stitch's implementation, and it looks like they are including both types in their implementation. Here's [the code I reviewed](https://github.com/singer-io/tap-github/blob/master/tap_github/__init__.py#L908), in case I'm reading it wrong.

What do you think? None of these are showstoppers in themselves and I'm curious of your thoughts here. I'm happy to discuss further here, or we can move that topic to another issue in order to speed along the addition of the PR stream and non-open issue states.

ericboucher:
@aaronsteers

I am not sure either. IMHO the fact that PRs are Issues from a GitHub API standpoint is a historical artifact. So I would be in favor of splitting them more cleanly.

However, from a "tap" standpoint, the goal is to get all the data as is, without too much interpretation. So from a "tap" perspective, it means getting all issues returned by the API, PRs included.

TLDR - I'll defer to your team as the main maintainers and make the changes needed either way :)

aaronsteers:
> @aaronsteers
> 
> I am not sure either. IMHO the fact that PRs are Issues from a GitHub API standpoint is a historical artifact. So I would be in favor of splitting them more cleanly.
> 
> However, from a "tap" standpoint, the goal is to get all the data as is, without too much interpretation. So from a "tap" perspective, it means getting all issues returned by the API, PRs included.
> 
> TLDR - I'll defer to your team as the main maintainers and make the changes needed either way :)

I can see going either way. @ericboucher - Have you checked if the PR stream has all the interesting/relevant fields in the Issues stream?

Would love feedback from other contributors @Ry-DS and @laurentS

ericboucher:
@aaronsteers diving a bit more into it, the PR stream includes more data for PRs as well as the fields included in 'issues'.

Therefore, I think my preference is to have the `Issues` stream pull issues only and not PRs, and the `PullRequest` stream gets PRs with additional info (such as base, target, ...)

Otherwise, it means that we would get the same object (PRs) twice to get all relevant information. Thoughts?

tap_github/streams.py
```diff
@@ -412,3 +423,254 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+        By default, will return context if provided and otherwise the record dict.
+        Developers may override this behavior to send specific information to child
+        streams for context.
+        """
+        if context is None:
+            raise ValueError("Issue stream should not have blank context.")
+
+        context["issue_number"] = record["number"]
+        context["comments"] = record["comments"]  # If zero, comments can be skipped
+        return context
```
aaronsteers, line 461: Since PullRequests don't have child stream types any more, I think this method is not used and can be safely removed. (Probably also not used for `IssuesStream` anymore but that's a separate topic.)

aaronsteers:
> @aaronsteers diving a bit more into it, the PR stream includes more data for PRs as well as the fields included in 'issues'.
> 
> Therefore, I think my preference is to have the `Issues` stream pull issues only and not PRs, and the `PullRequest` stream gets PRs with additional info (such as base, target, ...)
> 
> Otherwise, it means that we would get the same object (PRs) twice to get all relevant information. Thoughts?

@ericboucher 
Thanks for comparing those field lists. That's a big help.

The other thing I just realized though is that children - like issue comments - are generically labeled as "Issue Comments" and don't have any data in their record to tell the user whether the comment is actually an Issue comment or a PR comment. The join downstream then would get quite messy to get PR comments linked back up with PRs.

Instead of a simple left join to everything we need, like:

```sql
SELECT
    issues.id,
    issues.type,
    count(*) AS comment_count,
    max(issue_comments.post_date) as latest_comment,
    -- ...
FROM issue_comments
LEFT JOIN issues -- most things can be pull from here
ON ...
LEFT JOIN pull_requests. -- optional join, if PR-related fields are needed
ON ...
GROUP BY 1, 3
```

We'd have to re-union them back together in order to make the join, something like:

```sql
SELECT
    issues.id,
    issues.type,
    count(*) AS comment_count,
    max(issue_comments.post_date) as latest_comment
FROM issue_comments
LEFT JOIN (
    SELECT ... FROM issues 
    UNION ALL 
    SELECT ... FROM pull_requests
) AS all_issue_types
ON ...
GROUP BY 1, 3
```


In addition to the messiness of joins downstream, there are some other weird edge cases which I can imagine causing problems, like a user pulling just the issue_comments stream and issues streams (without selecting pull_requests) and then not having any means of knowing if the parent issue record is missing due to it being a PR or due to another data quality or replication issue.

So, I guess in summary, because children don't know what type of issue their parent is, I still would like to keep the issues list as a superset of all issues even though this is admittedly redundant.

I should add: there's an easy workaround to exclude PRs from the issues stream, which would be to use Stream Maps in the config and set something like:

```js
{
    // ..
    "stream_maps": {
        "issues": {
            "__filter__": "record['type'] = 'issue'"
        }
    }
{
```

Would you be okay keep as-is given the above?

ericboucher:
@aaronsteers works for me. I updated the streams following your input. Should I add the stream_map example in a comment somewhere maybe?

aaronsteers:
> @aaronsteers works for me. I updated the streams following your input. Should I add the stream_map example in a comment somewhere maybe?

Totally optional, but yes, if you think it could be helpful. Certainly it makes sense to call out Issues being inclusive also of Pull Requests. My little config snippet was written by hand from memory so it might take a little tweaking to get syntax just right. 
 😄

ericboucher:
@aaronsteers note that all the children's streams are also available for pulls. So we could also take the approach of getting their children separately. We will probably be getting commits and reviews.

https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps#pulls

aaronsteers:
> @aaronsteers note that all the `comments` stream is also available for pulls. So we could also take the approach of getting pulls children separately.
> 
> https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps#pulls

It appears so at first but I think not actually. The text if you read the description refers to "review comments" which are comments attached to a review. My understanding is that not all PR comments are "review comments".

This section from the docs points to the Issue Comments API for PR comments:

ericboucher:
Way to make this complicated, thanks GitHub! haha

ericboucher:
@aaronsteers I think we are good like this then. To summarize, these are the changes included in this PR:

- IssueStream pulls all issues and PRs regardless of state and we add a "`type` field making it easier to differentiate and filter
- PullRequestStream pulls all PRs regardless of state with the extra info available with the `/pulls` endpoint (base, target, ...)

tap_github/streams.py
```diff
@@ -231,6 +236,10 @@ def http_headers(self) -> dict:
         headers["Accept"] = "application/vnd.github.squirrel-girl-preview"
         return headers
 
+    def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
+        row["type"] = "pull_request" if row["pull_request"] else "issue"
```
aaronsteers, line 240: Looks like this is failing (missing?) for regular issues. What do you think of?:

```suggestion
        row["type"] = "pull_request" if "pull_request" in row else "issue"
```

aaronsteers (changes requested):

aaronsteers (approved):

aaronsteers:
@ericboucher - Do you mind resolving the conflicts from the recent merge to `main`? Otherwise, this looks ready to go. ✅

ericboucher:
@aaronsteers ready to go :)

PR #26, opened 2021-09-24
Add EventsStream

ericboucher:
Adding Events stream at the repository level


ericboucher:
@laurentS @aaronsteers 

Looks like we are hitting a limitation of the GitHub API on this one:

Request: `https://api.github.com/repos/octocat/Hello-World/events?per_page=1000&page=4&sort=updated`

```
{
  "message": "In order to keep the API fast for everyone, pagination is limited for this resource. Check the rel=last link relation in the Link response header to see how far back you can traverse.",
  "documentation_url": "https://docs.github.com/v3/#pagination"
}
```

It looks like we won't be able to traverse back in time very far for Events. Should we just find a way to stop after a few pages?

laurentS:
I tried `curl -vvv -H "Accept: application/vnd.github.v3+json"   "https://api.github.com/repos/octocat/Hello-World/events?per_page=1000&page=1&sort=updated"` now and got this header included in the response:

`link: <https://api.github.com/repositories/1296269/events?per_page=1000&page=2&sort=updated>; rel="next", <https://api.github.com/repositories/1296269/events?per_page=1000&page=3&sort=updated>; rel="last"`

On `facebook/react-native` (which I'd assume to be a lot more active than `Hello-World`, the last page allowed was 2.

So I suppose we could use one of the tap's methods to figure out how many pages we're allowed to fetch based on it. Then I guess it's the tap user's job to figure out how often to fetch this stream if they want full data coverage?

aaronsteers:
> So I suppose we could use one of the tap's methods to figure out how many pages we're allowed to fetch based on it. Then I guess it's the tap user's job to figure out how often to fetch this stream if they want full data coverage?

I think that is reasonable. If the tap knows how to paginate "as far back as possible" and "as possible" gets documented in the readme, this doesn't sound unreasonable. There are other data sources which have a similar max lookback window, and the guidance turns into "start early to capture a history of data" + "beware of letting the tap get disabled for longer than x days or you'll have gaps".

tap_github/client.py
```diff
@@ -64,11 +64,18 @@ def get_next_page_token(
...
             results = resp_json.get("items")
 
-        if results:
-            # Paginate as long as the response has items
-            return (previous_token or 1) + 1
+        # Paginate as long as the response has items
+        if not results:
+            return None
+
+        # Test if we have hit the last page for restricted pages such as "events".
```
ericboucher, line 71: Another option would be to rely on the `next` link directly instead of calculating `next_token` as `previous_token + 1`.

We could do something like this:
```
    next_url = response.links.get('next', {}).get('url')
    parsed = urlparse(url)
    captured_value_list = parse_qs(parsed.query).get('page')
    next_page_string = captured_value_list and captured_value_list[0]
    return int(next_page_string) if next_page_string and next_page_string.isdigit() else None
```

What do you think @aaronsteers ?

README.md
```diff
@@ -44,6 +43,16 @@ A small number of records may be pulled without an auth token. However, a Github
 
 ## Usage
 
+### API Limitation - Pagination
+
+The GitHub is limited for some resources such as `/events`. For some resources, users might encounter the following error:
```
aaronsteers, line 48:
```suggestion
The GitHub API is limited for some resources such as `/events`. For some resources, users might encounter the following error:
```

tap_github/client.py
```diff
@@ -58,15 +59,15 @@ def get_next_page_token(
         ):
             return None
 
-        resp_json = response.json()
-        if isinstance(resp_json, list):
-            results = resp_json
-        else:
-            results = resp_json.get("items")
+        # Leverage header links returned by the GitHub API.
```
ericboucher, line 62: @aaronsteers I revamped the next_token logic completely, let me know what you think.

ericboucher: I'm pretty confident GitHub sends the next link for all requests but I'll investigate a bit more!

ericboucher: https://docs.github.com/en/rest/guides/traversing-with-pagination

tap_github/streams.py
```diff
@@ -266,6 +266,112 @@ class CommunityProfileStream(GitHubStream):
...
+            return []
+
+        return super().get_records(context)
+
+    def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
+        # TODO - We should think about the best approach to handle this. An alternative would be to do
+        # a 'dumb' tap that just keeps the same schemas as GitHub without renaming these objects to "target_".
+        row["target_repo"] = row.pop("repo", None)
+        row["target_org"] = row.pop("org", None)
+        return row
```
aaronsteers, line 299: In doing a final pass, I came across this block and remembered in another PR we discussed there may be upcoming collisions on these two fields.

Are these redundant by chance with the `repo` and `org` string fields we are using for paritioning? Is there much to lose by removing them and just downstream, getting any additional needed properties like id/login/name from the parent stream using those parent context keys?

tap_github/streams.py
```diff
@@ -266,6 +266,112 @@ class CommunityProfileStream(GitHubStream):
...
+                th.Property("name", th.StringType),
+            ),
+        ),
+        th.Property(
+            "target_org",
+            th.ObjectType(
+                th.Property("id", th.StringType),
+                th.Property("login", th.StringType),
+            ),
+        ),
```
aaronsteers, line 328: Ditto question above re: whether these are redundant and could be removed with minimal issue.

aaronsteers (commented):
@ericboucher - I did a final pass and just had one last question regarding repo and org properties.

tap_github/streams.py
ericboucher: Sadly, when I checked these could be different. Esp for forks and merge events between forks and parents...

ericboucher: I added some details in the comment for now

ericboucher:
@aaronsteers ready to merge? :)

tap_github/repository_streams.py
```diff
@@ -278,6 +278,114 @@ class CommunityProfileStream(GitHubStream):
...
+                th.Property("url", th.StringType),
+            ),
+        ),
+        th.Property(
+            "payload",
+            th.ObjectType(
+                th.Property("before", th.StringType),
+                th.Property("action", th.StringType),
+                th.Property("comment", th.StringType),
+                th.Property("issue", th.StringType),
```
laurentS, line 360: at least on some events, `payload.issue` (and `payload.comment`) is a full-blown object, not just a string, for instance on `IssueCommentEvent`. Will this just stringify the content?
Same with `review` on `PullRequestReviewEvent` for instance.

laurentS (approved):
gtg in my mind, but worth checking the schema for different types of events.

tap_github/repository_streams.py
ericboucher: @aaronsteers do you know what the default behavior is in this case? Or what you would recommend doing instead?

edgarrmondragon: @laurentS @ericboucher 

> Will this just stringify the content?

I don't think so 🤔. It will just output either a string or an object, so targets downstream may fail during record-schema validation if an object is sent.

If the two record possibilities are

```json
{
  "issue": "<some-issue-id>",
  ...
}
```

```json
{
  "issue": {
    "id": "<some-issue-id>",
    ...
  },
  ...
}
```

then I'd recommend unifying them during `post_process` and declare them as a `th.ObjectType()`.

One option if the `object` overlaps with the `string` in on know  is to make this as `th.ObjectType()`, with

ericboucher: I expended the objects and I think the PR is ready for a final review.

@aaronsteers @edgarrmondragon are there any tools to extract the schema/typing directcly from an OpenAPI file? It would be pretty neat if we could simply plug into github/rest-api-description@main/descriptions-next/api.github.com/api.github.com.json to generate and validate types

edgarrmondragon: @ericboucher not any that do it as we need out-of-the-box, but I have an implementation here: https://github.com/edgarrmondragon/tap-shortcut/blob/99bb738608305baa11fe26acfe32d2c9c7089a58/tap_shortcut/tap.py#L86-L121.

Wdyt?

aaronsteers: I'd love to see if the OpenAPI spec can speed up SDK tap development. I've dropped an issue here: https://gitlab.com/meltano/sdk/-/issues/310

@ericboucher - Do you mind leaving a quick comment on that issue with your thoughts?

aaronsteers:
@edgarrmondragon and @ericboucher - For my part, this PR looks good to me. Approving. 👍 

@edgarrmondragon - You're a little deeper on some of the recent threads. Feel free to holler if you see any issues or blockers.

aaronsteers (approved):

edgarrmondragon (approved):
LGTM!

PR #27, opened 2021-09-27
Fix reactions mapping type in Issues and PRs streams

ericboucher:
Quick fix for reactions mapping. As a side question, now that we use the GitHub token, what do you think of adding more projects to tests @aaronsteers? We could propose https://github.com/mapswipe/mapswipe which @laurentS and I work on and has varied inputs.

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -357,19 +357,17 @@ class IssuesStream(GitHubStream):
         ),
         th.Property(
             "reactions",
-            th.ArrayType(
-                th.ObjectType(
-                    th.Property("url", th.StringType),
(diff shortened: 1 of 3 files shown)


aaronsteers:
> As a side question, now that we use the GitHub token, what do you think of adding more projects to tests @aaronsteers? We could propose https://github.com/mapswipe/mapswipe which @laurentS and I work on and has varied inputs.

@ericboucher  Sounds great!

aaronsteers:
LGTM! :shipit:

aaronsteers (approved):
Looks great. Thanks for adding the additional project to tests.

PR #29, opened 2021-10-08
Add ContributorsStream and AnonymousContributorsStream

ericboucher:
Add ContributorsStream

I think it's worth fetching anonymous contributors and not just GitHub ones, but it does add some complexity since we can't dedup by node_id. Let me know if you think we should keep it like that or maybe split these into two streams using filters? cc @laurentS @aaronsteers


laurentS (commented):
@ericboucher I added a couple of comments. I guess on the dedup aspect, github does the job of making sure that contributors beyond the 500 mark are not the same as the ones before (but on one example, I've spotted a user whose name appeared on both sides, so I assume github does not have the info to link all records, eg. if you push a commit with an email that is not linked to your github account).
On the config option aspect that I suggested, I'll let @aaronsteers provide guidance.

tap_github/streams.py
```diff
@@ -904,3 +904,44 @@ def http_headers(self) -> dict:
...
         ),
     ).to_dict()
+
+
+class ContributorsStream(GitHubStream):
+    """Defines 'Contributors' stream."""
+
+    name = "contributors"
+    path = "/repos/{org}/{repo}/contributors"
+    # primary_keys = ["id"]
```
laurentS, line 914:
```suggestion
    primary_keys = ["org", "repo", "node_id"]
```
Using `id` (which is the user id) leads to a bug where a single user can only have contributor data for one repo. In other words, if I've contributed to 2 repos for which I'm fetching this stream, I will only get 1 record for my contributions to one of them, the other record is missing in the output. The change above fixes that.

laurentS (changes requested):
There's a bug in the current stream implementation (see suggestion below).

tap_github/streams.py
ericboucher: We do still have the issue of the anonymous tap. One option I was thinking about was to split the anonymous contributors into another stream AnonymousContributorStream and filter it to ONLY grab anonymous contribution and the associated fields. What do you think?

ericboucher:
I split the contributor stream into two streams `Contributors` and `AnonymousContributors` because the fields are so different. One big caveat is that we are essentially querying contributor data twice, but sadly there doesn't seem to be a clean way to get anonymous contributions only.

In the future, I recommend disabling the anonymous stream by default and having a flag to activate it. OR find a smarter way to have the data land in two different tables while only running once with the `anon=true` flag activates.

tap_github/streams.py
```diff
@@ -938,10 +939,95 @@ class ContributorsStream(GitHubStream):
...
+                th.Property("id", th.IntegerType),
+                th.Property("node_id", th.StringType),
+                th.Property("avatar_url", th.StringType),
+                th.Property("gravatar_id", th.StringType),
+                th.Property("html_url", th.StringType),
+                th.Property("type", th.StringType),
+                th.Property("site_admin", th.BooleanType),
+            ),
+        ),
+    ).to_dict()
```
laurentS: This looks like a spillover from another PR, no?

tap_github/streams.py
```diff
@@ -938,10 +939,95 @@ class ContributorsStream(GitHubStream):
...
         th.Property("events_url", th.StringType),
         th.Property("received_events_url", th.StringType),
-        # Anonymous contributor keys. Only fetched with params["anon"] = 'true'
+        th.Property("type", th.StringType),
+        th.Property("site_admin", th.BooleanType),
+        th.Property("contributions", th.IntegerType),
+    ).to_dict()
+
+
+class AnonymousContributorsStream(GitHubStream):
```
laurentS, line 948: Could you define this as a child stream of `ContributorsStream`? Then you run the first stream, pass the number of non-anonymous contributors you received as some sort of context, and then run the child stream and start paginating from where the parent stream left off?
It's probably abusing the parent-child relationship a bit, and I don't know if there might be weird side effects as a result.

laurentS (commented):
Just added a thought about how to dedupe api requests...

tap_github/streams.py
ericboucher: Anonymous contributors are not necessarily at the end either... so I think that would be tricky.

tap_github/streams.py
ericboucher: Not sure on which commit # you got this 🤷 , probably a merge of main?

tap_github/streams.py
laurentS: arf, ok. Thanks github. Maybe there's a way to produce 2 different types of record messages for a single stream? But I don't know how :)

laurentS (approved):
LGTM!

PR #31, opened 2021-10-11
Workflow streams

edgarrmondragon:
New streams for GitHub Actions :

- [`workflows`](https://docs.github.com/en/rest/reference/actions#list-repository-workflows): All workflows  in a repository.
- [`workflow_runs`](https://docs.github.com/en/rest/reference/actions#list-workflow-runs-for-a-repository): All workflow runs  in a repository.
- [`workflow_run_jobs`](https://docs.github.com/en/rest/reference/actions#list-jobs-for-a-workflow-run): All jobs executed in a workflow run. For example, every value in a matrix corresponds to a job.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1420,7 +1420,7 @@ class WorkflowsStream(GitHubRestStream):
     primary_keys = ["id"]
     replication_key = "updated_at"
     parent_stream_type = RepositoryStream
-    ignore_parent_replication_key = False
+    ignore_parent_replication_key = True
     state_partitioning_keys = ["repo", "org"]
     records_jsonpath = "$.workflows[*]"
 
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1,7 +1,5 @@
 """Repository Stream types classes for tap-github."""
 
-import requests
-
 from typing import Any, Dict, Iterable, List, Optional
 
 import requests
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1520,7 +1520,6 @@ class WorkflowRunJobsStream(GitHubRestStream):
     name = "workflow_run_jobs"
     path = "/repos/{org}/{repo}/actions/runs/{run_id}/jobs"
     primary_keys = ["id"]
-    replication_key = "completed_at"
     parent_stream_type = WorkflowRunsStream
     ignore_parent_replication_key = False
     state_partitioning_keys = ["repo", "org", "run_id"]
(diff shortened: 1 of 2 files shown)


aaronsteers:
Exciting!

edgarrmondragon:
@aaronsteers Now just gotta work around the rate limiting 🙃

```
{"message":"API rate limit exceeded for installation ID 19254008.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

tap_github/streams.py
```diff
@@ -904,3 +907,171 @@ def http_headers(self) -> dict:
...
+        th.Property("labels", th.ArrayType(th.StringType)),
+        th.Property("runner_id", th.IntegerType),
+        th.Property("runner_name", th.StringType),
+        th.Property("runner_group_id", th.IntegerType),
+        th.Property("runner_group_name", th.StringType),
+    ).to_dict()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._schema_emitted = False
```
edgarrmondragon, line 1060: @aaronsteers I also made schema messages a bit quieter for jobs. It was outputting one every time a new parent workflow run was synced, and some targets flush records when they receive a new schema for an existing table, which would cause the target to write really small batches and affect performance.

Maybe this makes sense as a builtin option in the SDK?

ericboucher:
@edgarrmondragon is this still in progress? What's blocking?

edgarrmondragon:
> @edgarrmondragon is this still in progress? What's blocking?

@ericboucher thanks for the ping. I had forgotten about this one but last time tests were failing due to rate limits.

I just rebased and if tests are ✅ , this might be ready for review 😄

tap_github/repository_streams.py
```diff
@@ -1407,3 +1408,178 @@ def parse_response(self, response: requests.Response) -> Iterable[dict]:
...
+    """Defines 'workflows' stream."""
+
+    MAX_PER_PAGE = 100
+
+    name = "workflows"
+    path = "/repos/{org}/{repo}/actions/workflows"
+    primary_keys = ["id"]
+    replication_key = "updated_at"
+    parent_stream_type = RepositoryStream
+    ignore_parent_replication_key = False
```
ericboucher, line 1423: I think this should probably be True

edgarrmondragon: @ericboucher makes sense!

ericboucher (approved):

ericboucher:
Looks good to me! Same blocker as #93 for SonarCloud tho...

PR #33, opened 2021-10-18
Add StargazersStream

ericboucher:
Add StargazersStream.

Ready for review @laurentS @aaronsteers :)

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -940,7 +940,7 @@ class StargazersStream(GitHubStream):
         th.Property("org", th.StringType),
         th.Property("user_id", th.StringType),
         # Stargazer Info
-        th.Property("starred_at", th.IntegerType),
+        th.Property("starred_at", th.DateTimeType),
         th.Property(
             "user",
             th.ObjectType(


tap_github/streams.py
```diff
@@ -904,3 +904,46 @@ def http_headers(self) -> dict:
...
         ),
     ).to_dict()
+
+
+class StargazersStream(GitHubStream):
+    """Defines 'Stargazers' stream."""
+
+    name = "stargazers"
+    path = "/repos/{org}/{repo}/stargazers"
+    primary_keys = ["repo", "org"]
```
aaronsteers, line 914: Looks like this is missing `user_id` to make this unique. And, if correct, user.id would probably need to be aliased as a top-level `user_id` during `post_process()`.

tap_github/streams.py
```diff
@@ -904,3 +904,46 @@ def http_headers(self) -> dict:
...
+    parent_stream_type = RepositoryStream
+    state_partitioning_keys = ["repo", "org"]
+    replication_key = "starred_at"
```
aaronsteers, line 917: I'm not sure if this will capture removals, and if not, what do you think about disabling "incremental" on this stream by removing `replication_key`?

I'm not sure myself how to best handle this. In theory this is a relatively low-volume stream so full replication may be okay. But I can't say with confidence that the lack of deletion tracking is enough reason to fully remove incremental as an option.

Thoughts?

ericboucher: I think it can actually be a quite heavy stream as soon as you look at big projects like https://github.com/facebook/react-native with 100k stars...

Unstarring events are probably not that common too. Maybe we can add a warning and people can drop their table from time to time. Or do you have other reset/purge mechanisms?

tap_github/streams.py
```diff
@@ -904,3 +904,53 @@ def http_headers(self) -> dict:
...
+
+
+class StargazersStream(GitHubStream):
+    """Defines 'Stargazers' stream. Warning: this stream does NOT track star deletions."""
+
+    name = "stargazers"
+    path = "/repos/{org}/{repo}/stargazers"
+    primary_keys = ["repo", "org"]
+    parent_stream_type = RepositoryStream
+    state_partitioning_keys = ["repo", "org", "user_id"]
```
aaronsteers, line 916: Looks like these should be reversed. The primary key should be including user ID, but the state key (which drives how many bookmarks the stream will track) should be just ["repo", "org"].

Does that sound right?

```suggestion
    primary_keys = ["repo", "org", "user_id"]
    parent_stream_type = RepositoryStream
    state_partitioning_keys = ["repo", "org"]
```

aaronsteers (changes requested):

tap_github/streams.py
```diff
@@ -904,3 +904,53 @@ def http_headers(self) -> dict:
...
+        """
+        Add a user_id top-level field to be used as state replication key.
+        """
+        row["user_id"] = row["user"]["id"]
+        return row
+
+    schema = th.PropertiesList(
+        # Parent Keys
+        th.Property("repo", th.StringType),
+        th.Property("org", th.StringType),
```
aaronsteers, line 940: I think we're missing a top-level declare here on the grafted top-level key. 😄 

```suggestion
        th.Property("org", th.StringType),
        th.Property("user_id", th.StringType),
```

aaronsteers (approved):

aaronsteers:
> For CI maybe we should limit the `max` or `start_date` to avoid running into rate limits so often, What do you think?

Yeah, I'd be fine with either. I'm not sure what you mean by `max`, but I've used `start_date` for this before. The challenge with using `start_date` is that it doesn't accept relative references so we either need a script to populate it or we need to be periodically bumping it forward.

@ericboucher - What do you think of us removing 3.6 python check, since that's end of life in a few months anyway?

PR #37, opened 2021-10-27
Add dependents count fields on repositories stream

### LINKED ISSUE #34: Add dependents field on repositories stream
We use the `dependents` count for a repository, which is currently fetched by grabbing the html page for the project (eg. https://github.com/facebook/react/network/dependents) and parsing the HTML. As I write this ticket, the link above returns `7,878,702 Repositories` (and likewise for packages) and we grab these numbers.
Unfortunately, this info does not seem to exist anywhere in either the REST or graphQL APIs.

@aaronsteers Would you have any objection to me adding a request for that page to the `repositories` stream resulting in an extra field? Possibly behind some config option as it is fairly download heavy (the page above weighs 187kB).
Maybe in the `post_process` method? Ideally, the data will eventually be available in one of the APIs, and this can then be dropped.

laurentS:
This is WIP, as this commit requires changes in the SDK which are in a separate PR on that repo. Closes #34 

@aaronsteers following our conversation on the issue, I've taken a stab at this.
- This relies on a MR on the SDK to add `selected_by_default` to `Property`. I thought setting this to `false` for the 2 extra fields would prevent users from getting this extra data unless they explicitly ask for it by setting `selected: true` in the catalog.
- The test I wrote will fail until the sdk supports this properly. I opened [a merge request](https://gitlab.com/meltano/sdk/-/merge_requests/196) to help with fixing this, but it's WIP as well :see_no_evil: as I would rather have your opinion before spending more time on it
- If things work how I expect them to, then this would avoid the need for a config option like I initially imagined, and I think this would be much cleaner and also reusable across taps.

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -1,6 +1,8 @@
 """Stream type classes for tap-github."""
 
-from typing import Any, Dict, Iterable, List, Optional
+import re
+import requests
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 from singer_sdk import typing as th  # JSON Schema typing helpers
 
 from tap_github.client import GitHubStream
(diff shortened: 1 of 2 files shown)


ericboucher:
I agree with @aaronsteers that adding this as a stream will make it easier and not require the MR on the SDk.

How about 3 streams @laurentS:
1. get dependents and dependencies counts
2. get dependents repos
3. get dependency repos

https://github.com/dogsheep/github-to-sqlite/pull/70

PR #38, opened 2021-10-28
Add LanguagesStream

### LINKED ISSUE #36: Add RepoLanguagesStream
We would like the ability to fetch repo languages and LOC. The repo object has one main language, but a list is available under the languages_url with respective LOC - https://api.github.com/repos/facebook/react/languages

ericboucher:
Fixes #36

Adding languages stream.

A few annoying things:

1. GitHub returns an object of varying size {languageName1: size, languageName2: size, ...}.

To start, I went ahead and listed the top 100 or so languages (out of ~500). Another option would be to just create a much shorter/controlled list of our choosing. Or actually adding all of them. Any thoughts?

I used bigquery public datasets on repos to create this [list](https://docs.google.com/spreadsheets/d/14Bnc1QzLgKIxb9mH2W9FEplQOd-IP-0qK0zEdvKOhyQ/edit#gid=2090848317). And then found the config generating these languages on github - https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

2. A few languages (Fortran, Matlab and Pawn) seem to be written with two different case. Is there an easy way to make our parser case insensitive so that we only deal with one case? Looks like we can use the GitHub config to determine which one is the "normal" one.


laurentS:
Would it be easier to model the data something like this?
```
org: str
repo: str
language_name: str
lines: int
```
The `sdk`'s REST stream class has a [`parse_response`](https://gitlab.com/meltano/sdk/-/blob/main/singer_sdk/streams/rest.py#L352) method which you could overload to do just that.

ericboucher:
Great idea, I created an override for parse_response as you suggested. Let me know what you think @laurentS

tap_github/streams.py
```diff
@@ -270,17 +270,33 @@ class CommunityProfileStream(GitHubStream):
...
     name = "languages"
     path = "/repos/{org}/{repo}/languages"
-    primary_keys = ["repo", "org"]
+    primary_keys = ["repo", "org", "language_name"]
     parent_stream_type = RepositoryStream
     ignore_parent_replication_key = False
     state_partitioning_keys = ["repo", "org"]
 
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        """Parse the language response and reformat to return as an iterator of [{language_name: Python, byes: 23}]."""
```
laurentS, line 279:
```suggestion
        """Parse the language response and reformat to return as an iterator of [{language_name: Python, bytes: 23}]."""
```

tap_github/streams.py
```diff
@@ -270,17 +270,33 @@ class CommunityProfileStream(GitHubStream):
...
+            return []
+
+        languages_json = response.json()
+        results = [
+            {"language_name": key, "bytes": value}
+            for key, value in languages_json.items()
+        ]
+
+        for row in results:
+            yield row
```
laurentS, line 290:
```suggestion
        for key, value in languages_json.items()
            yield {"language_name": key, "bytes": value}
```
Just to save a bit of memory :)

laurentS (approved):
Just a typo, but otherwise looks good to me!

tap_github/streams.py
```diff
@@ -266,6 +267,34 @@ class CommunityProfileStream(GitHubStream):
...
+    state_partitioning_keys = ["repo", "org"]
+
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        """Parse the language response and reformat to return as an iterator of [{language_name: Python, bytes: 23}]."""
+        if response.status_code in self.tolerated_http_errors:
+            return []
+
+        languages_json = response.json()
+        for key, value in languages_json.items()
+            yield {"language_name": key, "bytes": value}
```
aaronsteers, line 285: Does this just return those languages in use or is it one row per item of everything here?:

https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

If the latter (nulls and zeros are present), we may want to do something like:

```suggestion
            if value:
                yield {"language_name": key, "bytes": value}
```

tap_github/streams.py
```diff
@@ -266,6 +267,34 @@ class CommunityProfileStream(GitHubStream):
...
+    ignore_parent_replication_key = False
+    state_partitioning_keys = ["repo", "org"]
+
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        """Parse the language response and reformat to return as an iterator of [{language_name: Python, bytes: 23}]."""
+        if response.status_code in self.tolerated_http_errors:
+            return []
+
+        languages_json = response.json()
+        for key, value in languages_json.items()
```
aaronsteers, line 284: I noticed the CI failure comes from this line. I think it just needs a colon.

```suggestion
        for key, value in languages_json.items():
```

tap_github/streams.py
laurentS: No, the tests we did on the endpoint only list the languages actually used in the repo, so it should be fine without filtering.

aaronsteers (approved):
Approving per the recent updates. Thanks!

ericboucher:
@aaronsteers let's ship it :)

PR #39, opened 2021-11-01
Add token rotation capability

ericboucher:
Add an authenticator to handle GitHub tokes and rotate keys to avoid hitting API limits.

Let me know what you think about this implementation @laurentS @aaronsteers


tap_github/client.py
```diff
@@ -110,6 +112,14 @@ def _request_with_backoff(
             )
             return response
 
+        # Rate limiting
+        if response.status_code == 403 and "Rate Limit Exceeded" in str(
+            response.content
+        ):
+            # Change token and force request retry.
+            self.github_authenticator.get_next_auth_token()
+            raise RuntimeError("Rate Limit Exceeded. Updating active token.")
```
laurentS, line 121: I don't understand these 2 lines. If you successfully rotate the token, you end up raising here, doesn't that interrupt the tap?

tap_github/autenticator.py
```diff
@@ -0,0 +1,102 @@
...
+from typing import Any, Dict, Optional
+
+from singer_sdk.streams import RESTStream
+
+
+class TokenRateLimit:
+    """A class to store token rate limiting information."""
+
+    DEFAULT_RATE_LIMIT = 5000
+    RATE_LIMIT_BUFFER = 1000
```
laurentS, line 16: Is this a buffer of quota you don't allow the tap to use? maybe worth a docstring?

laurentS (commented):
I'm obviously biased since we've discussed this before you wrote the code, but it does look helpful to me.
I think I would add a way to override the limit buffer value.
And I think there's a bug in the logic, see my comment.

tap_github/client.py
ericboucher: This is in a function that has a backoff mechanism, so it changes the token, then fails to force the function to rre-run. I'll add a note

tap_github/autenticator.py
```diff
@@ -13,15 +13,20 @@ class TokenRateLimit:
     """A class to store token rate limiting information."""
 
     DEFAULT_RATE_LIMIT = 5000
-    RATE_LIMIT_BUFFER = 1000
+    DEFAULT_RATE_LIMIT_BUFFER = 1000
 
-    def __init__(self, token: str):
+    def __init__(self, token: str, rate_limit_buffer: Optional[int]):
```
laurentS, line 21:
```suggestion
    def __init__(self, token: str, rate_limit_buffer: Optional[int] = None):
```

tap_github/autenticator.py
```diff
@@ -13,15 +13,20 @@ class TokenRateLimit:
...
         self.rate_limit_remaining = self.DEFAULT_RATE_LIMIT
         self.rate_limit_reset: Optional[int] = None
         self.rate_limit_used = 0
+        self.rate_limit_buffer = (
+            rate_limit_buffer
+            if rate_limit_buffer is not None
+            else self.DEFAULT_RATE_LIMIT_BUFFER
+        )
 
     def update_rate_limit(self, response_headers: Any):
```
laurentS, line 34:
```suggestion
    def update_rate_limit(self, response_headers: Any) -> bool:
```

ericboucher:
In the future, it would be nice to integrate directly on top of the existing authenticator property defined at the SDK level.
The use-case here is quite different from the existing functions and behavior, but it could be interesting to explore integrating a more generalizable version of this on top of `authenticators.APIAuthenticatorBase`

aaronsteers:
@ericboucher and @laurentS - Do I understand correctly that you want to pass multiple auth tokens in config, and have the tap automatically rotate at runtime among whatever keys are provided?

If yes, can you tell me a little more about how you are getting multiple auth keys and how you'd manage those? Is it the case where you'd combine multiple auth keys from members on a team, or perhaps combine the auth of one user account along with the auth of an app or repo key?

I'm also adding a comment inline that we can use the `description` property inline with the config to make this self-documenting.

tap_github/tap.py
```diff
@@ -26,7 +26,10 @@ class TapGitHub(Tap):
     config_jsonschema = th.PropertiesList(
         th.Property("user_agent", th.StringType),
         th.Property("metrics_log_level", th.StringType),
+        # Authorization options
         th.Property("auth_token", th.StringType),
+        th.Property("auth_tokens", th.ArrayType(th.StringType)),
+        th.Property("rate_limit_buffer", th.ArrayType(th.IntegerType)),
```
aaronsteers, line 32: SDK based taps and target can now provide inline documentation and then use that to print new documentation for a README.md file.

Here inline, for `auth_token` and `auth_tokens`, I think it would be helpful to note that they are mutually exclusive, and to describe the use cases for each.

```py
        th.Property("auth_token", th.StringType, description="..."),
        th.Property("auth_tokens", th.ArrayType(th.StringType), description="..."),
        th.Property("rate_limit_buffer", th.ArrayType(th.IntegerType), description="..."),
```

This makes them break into multi-line declarations once formatted, but the advantage to readability and inline expectation-setting is worth it, I think.

The readme docs for those settings can then be auto-generated with:

```
tap-github --about --format=markdown
```

aaronsteers (changes requested):

ericboucher:
So I think the code is ready from a functionality standpoint @aaronsteers 
The one thing to figure out is how to input the tokens:

1. where do we explain that one can use GITHUB_TOKEN* to pass a list of tokens? Simply in the readme?
2. should the different input options be incompatible or additive?
3. we could add the auth_token under auth_tokens to simplify the config, I didn't do that at first to keep the code backward compatible with existing configs people may have. But we are probably the main user of this tap atm so maybe I am overthinking this?

Thoughts @aaronsteers @laurentS ?

aaronsteers:
@ericboucher 

> 2. should the different input options be incompatible or additive?

My vote would be for one of these options:

**Option A: Mutually exclusive. If both are passed, an error is raised to explain expected usage.**

**Option B: Additive - which I think requires a setting rename to something like `additional_auth_tokens`.**

In the B option, the implied context that these will be rotated through if rate limits are met on the first passed token. I think this option would also be compatible with the main `auth_token` being null and all of them being passed as `additional_auth_tokens`. I think deduping between sets is also not a bad idea either in this scenario. (Up to you how resilient you want the handling to be. I can see value to allowing the user to just "pass all the auth keys I have access to" and for the tap to be smart enough to smartly handle duplicates without requiring the user to manually inspect them.)

> 1. where do we explain that one can use GITHUB_TOKEN* to pass a list of tokens? Simply in the readme?

I think the inline `description` is the best place, and then just copy-paste to the readme, or auto-generate the settings section of the readme using `tap-github --about --markdown=markdown`.

> 3. we could add the auth_token under auth_tokens to simplify the config, I didn't do that at first to keep the code backward compatible with existing configs people may have. But we are probably the main user of this tap atm so maybe I am overthinking this?

ericboucher:
@aaronsteers I went with option B, let me know what you think :)

laurentS:
Since this is hard to test in CI, I've been running the tap on 3 streams in parallel on my machine, with 2 tokens, one in `auth_token`, one in `additional_auth_tokens`. Things worked as expected:
- when a token hit its limit (minus buffer), the tap rotated to the next token
- when all tokens hit their limits, the tap stopped. Since I had 3 taps running in parallel, they all stopped roughly at the same time.
- The buffer is not *exact*, since parallel requests can go a little over the soft limit
- The only minor bug I could find is that the last request made, if it returns an HTTP 200, is still discarded, so there is potential for some data loss. I did not run any incremental streams, so I cannot confirm what happens with the state bookmark. A possible solution would be to check the validity of the token before making the request, so that the last valid request is actually saved.

ericboucher:
@laurentS good thinking, I made the change and now check for validity before the request instead of after

ericboucher:
@aaronsteers @laurentS ready to go :)

laurentS:
Just a heads up that this PR is broken right now. I'm trying to figure out where the problem is exactly, but I'm seeing error 403 from github with rate limiting messages, which we should not be getting if this was working correctly.

tap_github/client.py
```diff
@@ -106,8 +106,9 @@ def _request_with_backoff(
             return response
 
         # Rate limiting
-        if response.status_code == 403 and "Rate Limit Exceeded" in str(
-            response.content
+        if (
+            response.status_code == 403
+            and "rate limit exceeded" in str(response.content).lower()
```
ericboucher, line 111: nice job GitHub, changing the case sporadically haha

ericboucher:
PR is gtg now, thanks @laurentS for the lowercase fix

ericboucher:
Updated to the latest singer-sdk 0.3.17

laurentS:
You removed the check for "rate limit" in the headers on a 403. Is that on purpose? You can still get a "real" 403 for permissions, which should not be retried, I think.

ericboucher:
@laurentS great catch, re-added. It had gotten lost in the transition.

tap_github/client.py
```diff
@@ -85,55 +81,65 @@ def get_url_params(
...
+            FatalAPIError: If the request is not retriable.
+            RetriableAPIError: If the request is retriable.
+
+        .. _requests.Response:
+            https://docs.python-requests.org/en/latest/api/#requests.Response
+        """
         if response.status_code in self.tolerated_http_errors:
             self.logger.info(
-                "Request returned a tolerated error for {}".format(prepared_request.url)
+                "Request returned a tolerated error for {}".format(response.url)
```
laurentS, line 104: splitting hairs, but I'd rephrase that as `...tolerated status code...` just so that we don't have a million "errors" when grepping in logs :)

laurentS (approved):
just a tiny comment, but otherwise I think it's gtg :)

PR #42, opened 2021-11-09
Support python 3.10 in CI

laurentS:
The quotes are needed to prevent github from reading it as `3.1`.

### DIFF
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -10,7 +10,7 @@ jobs:
       GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3,10"]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


aaronsteers:
I'm worried out CI pipeline is failing regularly now due to rate limits. What do you think of in the short term, limiting to just two versions of python (maybe `3.8` and `3.10`) and then long-term try to get this in place: https://github.com/MeltanoLabs/tap-github/issues/43

laurentS:
I did not expect the CI to fail so much. That said, I did open 2 PRs within minutes, and I think Eric was pushing code at the same time, so we might have made things worse without realising.

Anyway, we're running on python 3.9, so I'm biased towards keeping it in CI for now. Adding 3.10 is still a nice to have, but a step towards us switching to it. If it was down to me, I'd leave this PR as is until we've resolved the rate limiting issue. Happy to limit to 3.8+3.9 as a temp solution if it works for you.

.github/workflows/test_tap.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
```
aaronsteers, line 13:
```suggestion
        python-version: [3.7, 3.8, 3.9, "3.10"]
```

aaronsteers (approved):

PR #46, opened 2021-11-10
Clean control chars from issue body field

### LINKED ISSUE #44: Target fails with chr(0) in issue body
I've run into an annoying bug with [target-postgres](https://github.com/datamill-co/target-postgres/) (datamill variant) where it crashes because a unicode char \x000 is sent to postgresql, causing it to reject the data.

The reason I'm reporting this here is that the data comes from this tap, and I'm not entirely sure what to do about it.

To reproduce:
- run `tap-github` to fetch this issue: https://github.com/cockroachdb/cockroach/issues/68703 (you can get to the bug faster by hardcoding the issue number in the `IssuesStream` url, so as to only get this one issue)
- The body of this issue contains a weird character (see screenshot from github below) which curl returns as `\u0000` and python requests as `\x00` (this might be the first time I write an issue report which might cause the bug to happen again :exploding_head: )
image
- Through the various encoding/decoding, this char ends up being sent to postgresql which chokes on it, and thows `psycopg2.errors.CharacterNotInRepertoire: invalid byte sequence for encoding "UTF8": 0x00`

On the one hand, I feel like the target should clean the data before sending it to postgres, but the same reasoning can apply to the tap. I'm not sure what the best practice is here. The interface is JSON, and I found no evidence that `\x00` is illegal in JSON, as long as it's escaped.

Any thoughts?

laurentS:
Fixes #44 

I've run the stream on just over 200k issues to test and have not seen any problems so far (including the one that was causing the problem in the first place).
The check for an empty body puzzles me a bit, but the tap failed almost immediately without it.

### DIFF
--- a/tap_github/streams.py
+++ b/tap_github/streams.py
@@ -310,12 +310,12 @@ class IssuesStream(GitHubStream):
 
     def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
         row["type"] = "pull_request" if "pull_request" in row else "issue"
-        if row['body'] is not None:
+        if row["body"] is not None:
             # some issue bodies include control characters such as \x00
             # that some targets (such as postgresql) choke on. This ensures
             # such chars are removed from the data before we pass it on to
             # the target
-            row['body'] = row['body'].encode('utf-8', errors='ignore')
+            row["body"] = row["body"].encode("utf-8", errors="ignore")
         return row
 
     schema = th.PropertiesList(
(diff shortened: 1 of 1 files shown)


ericboucher (approved):

aaronsteers:
@ericboucher - Approved - thanks for the detail around tests run. Are you able to merge once action items and approvals are in? If not, just lmk when this is ready to go and I'll merge with your blessing. Thanks!

laurentS:
Nope, I'm not allowed to merge, but happy for you to do it :)

PR #50, opened 2021-11-23
Fix deocoding bug with issues and issue comments

ericboucher:
- Also fix this bug for issue comments and pull requests
- Re-decode after encoding without errors


edgarrmondragon (approved):
lgtm 👍

PR #53, opened 2021-11-24
Adjust max-per-page constant to match api docs

laurentS:
Trivial PR to correct a typo in the code.
A quick review of github API docs (eg [here](https://docs.github.com/en/rest/reference/repos#list-organization-repositories)) says that the max is 100, not 1k.

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -10,7 +10,7 @@ from singer_sdk.streams import RESTStream
 class GitHubStream(RESTStream):
     """GitHub stream class."""
 
-    MAX_PER_PAGE = 1000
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
@laurentS I think this is addressed in https://github.com/MeltanoLabs/tap-github/pull/30 ([diff](https://github.com/MeltanoLabs/tap-github/pull/30/files#diff-fda9caa050a8af612a9b9298d6212c7dcbc59b037ad1e2198fce3e1ecf040288R13))

aaronsteers:
As #30 is more involved so I opted to go ahead and merge this as a quick fix. 👍

PR #55, opened 2021-11-25
Add ReadmeHtmlStream

ericboucher:
Add a new stream `ReadmeHtmlStream` to get the html version of the readme.


tap_github/streams.py
```diff
@@ -196,6 +197,47 @@ class ReadmeStream(GitHubStream):
     ).to_dict()
 
 
+class ReadmeHtmlStream(GitHubStream):
```
laurentS, line 200: It might be useful to add a docstring to explain why there seems to be 2 streams for README :)

ericboucher: done

aaronsteers:
@ericboucher - Do you think it could it be feasible to add this as a property of the existing `ReadmeStream`? Understanding we still need to make an extra http call to get the html version, it feels like from a user perspective, it could be simpler and more efficient to manage this as a single stream. What do you think?

edgarrmondragon:
@aaronsteers Perhaps this could work as a partition of `ReadmeStream`, if https://gitlab.com/meltano/sdk/-/issues/273 is addressed?

ericboucher:
@aaronsteers I agree that it could be nice from a user perspective but I don't see an easy way to do that with the current implementation of the tap. Making a second request feels like an anti-pattern here.

We would probably have to overwrite `prepare_request` and `request_records` while also changing the HTTP header setting in the middle. If you have an idea of an easier way to do this let me know and I'm happy to give it a shot :)

Otherwise, I'm tempted to merge this as is until we have https://gitlab.com/meltano/sdk/-/issues/273 for example to make it a bit easier :)

tap_github/repository_streams.py
```diff
@@ -153,6 +153,12 @@ def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
 
 
 class ReadmeStream(GitHubStream):
+    """
+    A stream dedicated to fetching the object version of a README.md.
+
+    Inclduding its content, base64 encoded. For html, see ReadmeHtmlStream.
```
laurentS, line 165: maybe worth clarifying that this content is github-flavoured-markdown as opposed to the html below?

laurentS (approved):
Looks good to me. I wonder if there would be a way to have either of these not selected_by_default (see https://gitlab.com/meltano/sdk/-/merge_requests/196) at some point. Either way, it's probably fine to merge now :)

PR #56, opened 2021-12-12
Correct property type for users stream

laurentS:
This PR fixes the data type for user logins, which clearly isn't an integer :)


aaronsteers:
Lgtm!

ericboucher (approved):

Issue #43, opened 2021-11-09
Add `requests-cache` implementation in CI to eliminate rate limit failures


aaronsteers:
@edgarrmondragon had an initial working implementation of requests caching using the `cached-requests` library: https://requests-cache.readthedocs.io/en/stable/

Used here: https://github.com/MeltanoLabs/tap-stackexchange/blob/main/tap_stackexchange/client.py#L7

And we have an SDK discussion on this here: https://gitlab.com/meltano/sdk/-/issues/237

@laurentS, @edgarrmondragon, and @ericboucher - If ever there were a good use case for this, I think this repo is a great one.

What do you all think of this approach?

1. When running tests in CI, invoke the `requests-cache` library to store http responses.
2. Sequence the build so that one python version is the main one (`3.9`, perhaps), and the main one runs prior to the others and stores its artifacts in a GitHub CI cache.
3. The other python versions run after the main version.
4. Even the main python version recalls the cache and thereby only has to send new requests not covered by a prior build's requests. (For instance, when a new stream is added.)

At most, this means one "main" job needs to send http requests - but in practice, even the main job should not need to make the full set of requests.

Optionally, a scheduled job could kill the cache once a week and refresh from scratch.

laurentS:
Relates to #14 
I think the approach you suggest makes sense. I was pondering using a sort of snapshot system saved in the repo (I think vcr.py does that), but using a cache like you suggest is probably less maintenance.

Github's API rate limiting has a 1h window, after which it resets, so we could have a daily job, or even twice daily scheduled job that busts/primes the cache. I'm thinking of what would happen if github's API format suddenly changes in a way that breaks the tap, the less time we wait for CI to fail the better. If the scheduled job refreshes the caches and then runs the CI job on it, we'd get notified of failures quickly. In that case, we can even run all "regular" CI jobs from cache, so they would be much faster as well, which never hurts (and there's still the risk that 2-3 eager devs are pushing code multiple times within a 1h window and hit limits anyway :wink: )

aaronsteers:
> I was pondering using a sort of snapshot system saved in the repo (I think vcr.py does that), but using a cache like you suggest is probably less maintenance.

I agree. In the related SDK issue, we imagine a dev lifecycle where eventually (if not in the first iteration due to PII), the caches are stored in yaml within the repo. This can power CI tests (except when we intentionally ignore/refresh the cache) gives us `--demo` behavior down the road without having to require creds that connect to the source.

Issue #14, opened 2021-09-12
Should the tap tests mocks github's API responses?


laurentS:
While developing I've run a few times into github's api rate limits (with my auth_token set!), which gets pretty annoying, as the tap does not work at all then.
This might become an issue for CI as well when the number of streams increases (the limit without an auth token is 60 requests/hour per IP).
So I am wondering if we should setup some sort of mocking of API calls, for instance using pytest's [monkeypatching ](https://docs.pytest.org/en/latest/how-to/monkeypatch.html#monkeypatching-returned-objects-building-mock-classes) fixture, or some other mechanism.

Pros:
- faster and more reliable test suite (so probably more tests written)
- better development experience (running tests would be almost instant)

Cons:
- it takes a little time to build the mock responses
- if the actual API changes its format, the mocks need to be updated. Tools like [VCR.py](https://vcrpy.readthedocs.io/en/latest/) could be helpful with this.
- breaking changes in the actual API would not break the tap's tests, but something downstream would probably fail. Unless we run a periodic CI job to compare mocks to actual API output.

aaronsteers:
@laurentS - I love this idea and I'm glad you referenced the VCR.py library. Do you mind adding some thoughts here: https://gitlab.com/meltano/sdk/-/issues/30#note_677520346

1. Yes - I think this kind of a mock capability would be fantastic.
2. If we can find a way to build the capability generically into the SDK, all the better. (Even if that is done in a future phase 2.)

aaronsteers:
> While developing I've run a few times into github's api rate limits (with my auth_token set!), which gets pretty annoying, as the tap does not work at all then.

@laurentS - Less exciting of a solution, but as a stop-gap, we could also limit the CI pipeline to only run full tests on python 3.9. Currently we're running full tests on all 4 supported python versions. We could set the other three to run partial tests only. Again, not ideal.

laurentS:
> > While developing I've run a few times into github's api rate limits (with my auth_token set!), which gets pretty annoying, as the tap does not work at all then.
> 
> @laurentS - Less exciting of a solution, but as a stop-gap, we could also limit the CI pipeline to only run full tests on python 3.9. Currently we're running full tests on all 4 supported python versions. We could set the other three to run partial tests only. Again, not ideal.

[...]

PR #60, opened 2021-12-19
Cache api requests in tests

laurentS:
Relates to #43 and #14

This PR sets up a simple caching mechanism for github api requests made by the tap during testing.

:warning: I had to make this PR on top of #59 because dependency resolution wouldn't complete otherwise.

The implementation is pretty much as suggested by @aaronsteers in #43:
- the matrix is setup to run jobs sequentially (specifically, not more than 1 job in parallel :wink: ) With caching, on my machine, `pytest` runs in 7 sec instead of 125s, so doing jobs in sequence does not really matter.
- the api responses are saved in a sqlite db file, which github should cache after a job completes successfully.
- github keeps cached files for up to 7 days (more if they keep being accessed), but we expire the content of a response after 24h.
- API keys should be excluded from the cached files

In practice, this means that a job (first or other in the matrix) will check the cache before hitting the API, and only make the actual request if it hasn't been sent in the past 24h.

The implementation is pretty simple, but should include all requests (except ones returning error codes), and also work locally for development, which is a nice bonus!

### DIFF
--- a/tap_github/tests/__init__.py
+++ b/tap_github/tests/__init__.py
@@ -1,1 +1,19 @@
 """Test suite for tap-github."""
+import requests
+import requests_cache
+
+# Setup caching for all api calls done through `requests` in order to limit
+# rate limiting problems with github.
+# Use the sqlite backend as it's the default option and seems to be best supported.
+# To clear the cache, just delete the sqlite db file at api_calls_tests_cache.sqlite
+# in the root of this repository
+requests_cache.install_cache(
+    "api_calls_tests_cache",
+    backend="sqlite",
+    # make sure that API keys don't end up being cached
+    ignored_parameters=["Authorization"],
+    # tell requests_cache to check headers for the above parameter
+    match_headers=True,
+    # expire the cache after 24h (86400 seconds)
+    expire_after=24 * 60 * 60,
+)
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -23,7 +23,7 @@ jobs:
         path: ./api_calls_tests_cache.sqlite
         # github cache expires after 1wk, and we expire the content after 24h
         # this key should not need to change unless we need to clear the cache
-        key: api-cache-v1
+        key: api-cache-v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -10,7 +10,7 @@ jobs:
       GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -11,9 +11,19 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9]
+      # run the matrix jobs one after the other so they can benefit from caching
+      max-parallel: 1
 
     steps:
     - uses: actions/checkout@v2
+    - name: Cache github API responses
+      uses: actions/cache@v2
+      with:
+        # must match the path in tests/__init__.py
+        path: ./api_calls_tests_cache.sqlite
+        # github cache expires after 1wk, and we expire the content after 24h
+        # this key should not need to change unless we need to clear the cache
+        key: api-cache-v1
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+api_calls_tests_cache.sqlite
 
 # Translations
 *.mo
--- a/poetry.lock
+++ b/poetry.lock
@@ -30,7 +30,7 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "black"
-version = "21.9b0"
+version = "21.12b0"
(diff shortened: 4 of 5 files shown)


tap_github/tests/__init__.py
```diff
@@ -1 +1,19 @@
...
+requests_cache.install_cache(
+    "api_calls_tests_cache",
+    backend="sqlite",
+    # make sure that API keys don't end up being cached
+    ignored_parameters=["Authorization"],
+    # tell requests_cache to check headers for the above parameter
+    match_headers=True,
+    # expire the cache after 24h (86400 seconds)
+    expire_after=24 * 60 * 60,
+)
```
aaronsteers, line 19: @laurentS - This is great; I'm very impressed how little code is needed! Also, I didn't know exactly how "ignored_parameters" would turn out but "Authorization" is a standard header which also seems a (somewhat) portable solution.

Did you scan the SQLite DB contents to see if there were any other auth or sensitive tokens captured?

Would be great to see this added to the SDK's cookiecutter - or perhaps into the `Code Samples` page on the docs.

cc @edgarrmondragon 

PS - Meltano is officially OOO the next two weeks so please pardon any delays in communications during the meanwhile. 😄

aaronsteers (approved):

ericboucher (approved):

tap_github/tests/__init__.py
laurentS: Nothing that looks problematic, no. Here's an example of the content for one entry taken from the sqlite database (after I changed the serializer from `pickle` to `json` for readability :sweat_smile:). Most importantly, the `Authorization` header is stripped as expected. So I think it's safe to cache this.

As for adding this into the SDK, I think pushing this to code samples would make more sense, as I don't know how generalisable the config is.

```json
{
  "content": "<content of the response, encoded>"
  "created_at": "2021-12-20T15:11:43.947734",
  "elapsed": 0.295303,
  "encoding": "utf-8",
  "expires": "2021-12-21T15:11:43.947679",
  "headers": {
    "Server": "GitHub.com",
    "Date": "Mon, 20 Dec 2021 15:11:43 GMT",
    "Content-Type": "application/json; charset=utf-8",
    "Transfer-Encoding": "chunked",
    "Cache-Control": "private, max-age=60, s-maxage=60",
    "Vary": "Accept, Authorization, Cookie, X-GitHub-OTP, Accept-Encoding, Accept, X-Requested-With",
    "ETag": "W/\"d6a3e7fa8493843d598ab6fecb3775a3eb80e77269baf7425371d0c231b43ab4\"",
    "X-OAuth-Scopes": "",
    "X-Accepted-OAuth-Scopes": "",
    "X-GitHub-Media-Type": "github.v3; param=star; format=json",
    "Link": "<https://api.github.com/repositories/1296269/stargazers?per_page=100&page=12&sort=updated&direction=asc&since=2021-12-20T00%3A00%3A00%2B00%3A00>; rel=\"prev\", <https://api.github.com/repositories/1296269/stargazers?per_page=100&page=14&sort=updated&direction=asc&since=2021-12-20T00%3A00%3A00%2B00%3A00>; rel=\"next\", <https://api.github.com/repositories/1296269/stargazers?per_page=100&page=18&sort=updated&dir
ection=asc&since=2021-12-20T00%3A00%3A00%2B00%3A00>; rel=\"last\", <https://api.github.com/repositories/1296269/stargazers?per_page=100&page=1&sort=updated&direction=asc&since=2021-12-20T00%3A00%3A00%2B00%3A00>; rel=\"first\"",
    "X-RateLimit-Limit": "5000",
    "X-RateLimit-Remaining": "921",
    "X-RateLimit-Reset": "1640015555",
    "X-RateLimit-Used": "4079",
    "X-RateLimit-Resource": "core",
    "Access-Control-Expose-Headers": "ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset",
    "Access-Control-Allow-Origin": "*",
    "Strict-Transport-Security": "max-age=31536000; includeSubdomains; preload",
    "X-Frame-Options": "deny",
    "X-Content-Type-Options": "nosniff",
    "X-XSS-Protection": "0",
    "Referrer-Policy": "origin-when-cross-origin, strict-origin-when-cross-origin",
    "Content-Security-Policy": "default-src 'none'",
    "Content-Encoding": "gzip",
    "X-GitHub-Request-Id": "DE0C:CC89:1760597:1832969:61C09D2F"
  },
  "raw": {
    "decode_content": false,
    "reason": "OK",
    "status": 200,
    "version": 11
  },
  "reason": "OK",
  "request": {
    "body": "PH%2y",
    "headers": {
      "User-Agent": "python-requests/2.26.0",
      "Accept-Encoding": "gzip, deflate",
      "Accept": "application/vnd.github.v3.star+json",
      "Connection": "keep-alive"
    },
    "method": "GET",
    "url": "https://api.github.com/repos/octocat/Hello-World/stargazers?per_page=100&page=13&sort=updated&direction=asc&since=2021-12-20T00%3A00%3A00%2B00%3A00"
  },
  "status_code": 200,
  "url": "https://api.github.com/repos/octocat/Hello-World/stargazers?per_page=100&page=13&sort=updated&direction=asc&since=2021-12-20T00%3A00%3A00%2B00%3A00"
}
```

laurentS:
The job just failed with an error that I find no proper docs about: `api rate limit exceeded for installation id 123456`. Others mention a similar problem with CI runs, it seems to be a limit per machine/IP. We can't do much about this, except that it highlights that the cache is not being used, at least not on all requests.

Also weirdly, the cached file weighs 2MB according to github, but on my machine it's closer to 60MB for the same test suite.

laurentS:
I'm going to merge this as is. The last run after clearing the cache shows that the test suite runs in 1min24s for python 3.7 (there is no cache available), and then 22 and 26s for python 3.8/3.9. I'll take this as a hint that caching actually works, although on my laptop the speed up is much more dramatic and the test suite completes in 7s with caching.

PR #63, opened 2021-12-22
Cache poetry dependencies

ericboucher:


laurentS (commented):
Nice! Maybe take the opportunity to update poetry to latest version?

edgarrmondragon:
@ericboucher This is a nice addition! Does this mean that all jobs running for the same commit sha will reuse the cached `.venv` and any new commit will invalidate cache?

ericboucher:
All commits with the same poetry.lock will use the same cache. Note that to be safe we create one cache per Python version to avoid edge cases

laurentS (approved):
LGTM!

PR #65, opened 2022-01-05
chore: Add pre-commit

### LINKED ISSUE #61: Add precommit hook for mypy and black
Add a simple pre-commit hook for `mypy` and `black` to avoid pushing code with errors that are quick and easy to fix.

edgarrmondragon:
Add pre-commit configuration for

- black
- mypy
- isort
- pyugrade (to ensure "newer" Python syntax is used)
- misc. checks (yaml, toml, json, whitespace)

Configuration for black, isort and mypy lives in `pyproject.toml`.

Closes #61


edgarrmondragon:
I wonder if we should add a https://github.com/marketplace/pre-commit-ci job

pyproject.toml
```diff
@@ -22,6 +22,20 @@ black = {version = "^21.5b1", allow-prereleases = true}
 types-requests = "^2.25.6"
 requests-cache = "^0.8.1"
 
+[[tool.mypy.overrides]]
+module = [
+    "backoff",
+]
+ignore_missing_imports = true
```
ericboucher, line 29: just curious, why is this needed?

ericboucher (approved):

ericboucher:
> I wonder if we should add a [github.com/marketplace/pre-commit-ci](https://github.com/marketplace/pre-commit-ci) job

@edgarrmondragon is the pre-commit making the changes locally on fails? So that the dev just has to commit the changes to fix? I think that would be enough rather than having a pre-commit-ci :)

pyproject.toml
edgarrmondragon: @ericboucher when a package like `backoff` is missing library stubs (such as `types-requests` for `requests`) or a `py.typed` file, mypy cannot get typing information from it.

This is the same as putting `# type: ignore` in front of any `import backoff` lines to make mypy happy.

edgarrmondragon:
@ericboucher 

 > @edgarrmondragon is the pre-commit making the changes locally on fails?

That's correct. If a user runs `pre-commit install` then all the checks described above will run on every commit locally.

pre-commit-ci as a pipeline fails when pre-commit would make changes, so it's a way to not have users install pre-commit locally.

pyproject.toml
ericboucher: So written like this, it only gets applied to the "backoff" module?

ericboucher:
Let's start with this approach. You can go ahead and merge if the following comment is true - https://github.com/MeltanoLabs/tap-github/pull/65#discussion_r782447728

PR #66, opened 2022-01-11
Add users StarredStream

ericboucher:
Adding a stream to get users starred projects.

For this stream as well as the stream for repos, I was thinking of implementing an early stop feature in the `get_next_page_token` function because the GitHub API does not provide a `since=` parameter for these and the number of `Stargazers` can get really big for famous projects.

My initial idea is exposed below, unfortunately, we don't have context in this function. What would you guys recommend as a workaround? @laurentS @edgarrmondragon 

```
if self.replication_key and self.replication_key in results[-1]:
    if results[-1][self.replication_key] > self.self.get_starting_timestamp(context):
      return None
```

I ended up using the `since` flag from the request params. Let me know what you think.


laurentS (approved):
Just spotted one type error.
It's a bit sad that this single endpoint does not work like everything else.

ericboucher:
Quick question @aaronsteers @edgarrmondragon  
What is the behavior for `ignore_parent_replication_key`? Should star streams have it set to `True` you think?

edgarrmondragon:
@ericboucher 

> What is the behavior for `ignore_parent_replication_key`? Should star streams have it set to `True` you think?

I think so. I just checked that "starring" doesn't affect the `updated_at` value for my own user object 👍

tap_github/user_streams.py
```diff
@@ -74,3 +74,86 @@ def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
...
+                    "owner",
+                    th.ObjectType(
+                        th.Property("login", th.StringType),
+                        th.Property("id", th.IntegerType),
+                        th.Property("node_id", th.StringType),
+                        th.Property("type", th.StringType),
+                        th.Property("avatar_url", th.StringType),
+                        th.Property("html_url", th.StringType),
+                        th.Property("site_admin", th.BooleanType),
+                    ),
```
edgarrmondragon, line 132: Probably outside of scope for this PR, but this schema are reused [in a few places](https://github.com/MeltanoLabs/tap-github/blob/4b6d7ca9ba604739b1ee3615acd84ea7c5c67b7b/tap_github/repository_streams.py#L89-L98), so perhaps an `OWNER_SCHEMA` can be refactored out of this?

Other schemas are likely reused too.

edgarrmondragon (approved):

tap_github/user_streams.py
ericboucher: Yeah I agree, it would be nice to have a file with shared schemas. I'll open an issue.

PR #69, opened 2022-01-21
Hot fix - StargazersSstream - user_id is an integer

ericboucher:


ericboucher:
Merged as part of #26

PR #70, opened 2022-01-26
Override --test behaviour to get schema messages (temp solution)

ericboucher:
This is a temporary fix to populate schemas without actually running the streams.

It will be rendered unnecessary once https://gitlab.com/meltano/sdk/-/merge_requests/218 gets merged but is a useful addition.


aaronsteers:
Hi, @ericboucher - Did we get a resolution with the closure of https://gitlab.com/meltano/sdk/-/merge_requests/218?

Is this workaround still needed?

ericboucher:
@aaronsteers we were waiting for the release. But I think it should be good now. cc @laurentS

ericboucher:
Not necessary once #77 is merged

PR #71, opened 2022-01-26
Add StatsContributorsStream

ericboucher:
Add the repo `/stats/contributors` endpoint to fetch weekly contribution data

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1331,7 +1331,7 @@ class StatsContributorsStream(GitHubStream):
             "a": "additions",
             "c": "commits",
             "d": "deletions",
-            "w": "week_start"
+            "w": "week_start",
         }
         parsed_response = super().parse_response(response)
         for contributor_activity in parsed_response:
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1341,7 +1341,6 @@ class StatsContributorsStream(GitHubStream):
                 if sum(week[key] for key in ["a", "c", "d"]) == 0:
                     continue
                 week_with_author = {replacement_keys[k]: v for k, v in week.items()}
-                week_with_author.update(week)
                 week_with_author.update(contributor_activity["author"])
                 week_with_author["user_id"] = week_with_author.pop("id")
                 yield week_with_author
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1327,6 +1327,12 @@ class StatsContributorsStream(GitHubStream):
 
     def parse_response(self, response: requests.Response) -> Iterable[dict]:
         """Parse the response and return an iterator of flattened contributor activity."""
+        replacement_keys = {
+            "a": "additions",
+            "c": "commits",
+            "d": "deletions",
+            "w": "week_start"
+        }
         parsed_response = super().parse_response(response)
         for contributor_activity in parsed_response:
             weekly_data = contributor_activity["weeks"]
@@ -1334,7 +1340,7 @@ class StatsContributorsStream(GitHubStream):
                 # no need to save weeks with no contributions.
                 if sum(week[key] for key in ["a", "c", "d"]) == 0:
                     continue
-                week_with_author = dict()
+                week_with_author = {replacement_keys[k]: v for k, v in week.items()}
                 week_with_author.update(week)
                 week_with_author.update(contributor_activity["author"])
                 week_with_author["user_id"] = week_with_author.pop("id")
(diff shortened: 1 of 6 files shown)


tap_github/repository_streams.py
```diff
@@ -1307,3 +1307,55 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+
+    name = "stats_contributors"
+    path = "/repos/{org}/{repo}/stats/contributors"
+    primary_keys = ["org", "repo", "user_id"]
+    parent_stream_type = RepositoryStream
+    ignore_parent_replication_key = False
+    state_partitioning_keys = ["repo", "org"]
+    # Note - these queries are expensive and the API might return an HTTP 202 if the response
+    # has not been cached recently. https://docs.github.com/en/rest/reference/metrics#a-word-about-caching
+    tolerated_http_errors = [202]
```
laurentS, line 1326: @edgarrmondragon the docs say that the request should be retried "a bit later" when we get a 202. Is there some sort of mechanism in the sdk to do this (I'd imagine something like enqueueing retries after N seconds)? I guess it could be handle the same way as a `429 - Too many requests` but that would less than ideal.

tap_github/repository_streams.py
```diff
@@ -1307,3 +1307,55 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        """Parse the response and return an iterator of flattened contributor activity."""
+        parsed_response = super().parse_response(response)
+        for contributor_activity in parsed_response:
+            weekly_data = contributor_activity["weeks"]
+            for week in weekly_data:
+                # no need to save weeks with no contributions.
+                is_week_empty = sum(week[key] for key in ["a", "c", "d"]) > 0
```
laurentS, line 1335: How do we distinguish a missing response (say if we received a 202 that has not been retried yet, something like `undefined`) from this situation where it's a `zero`, in the downstream processing?

tap_github/repository_streams.py
```diff
@@ -1307,3 +1307,55 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+                week_with_author.update(contributor_activity["author"])
+                yield week_with_author
+
+    schema = th.PropertiesList(
+        # Parent keys
+        th.Property("repo", th.StringType),
+        th.Property("org", th.StringType),
+        th.Property("user_id", th.IntegerType),
+        # Activity keys
+        th.Property("w", th.IntegerType),
```
laurentS, line 1349: I'd be tempted to rename these fields so they're more or less self-documenting:
- `week_start`
- `additions`
- `deletions`
- `commits`

What do you think?

tap_github/repository_streams.py
```diff
@@ -1307,3 +1307,55 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+        th.Property("a", th.IntegerType),
+        th.Property("d", th.IntegerType),
+        th.Property("c", th.IntegerType),
+        # Contributor keys
+        th.Property("login", th.StringType),
+        th.Property("id", th.IntegerType),
```
laurentS, line 1355: Is `id` always the same as `user_id` here?

tap_github/repository_streams.py
```diff
@@ -1307,3 +1307,55 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+                week_with_author = dict()
+                week_with_author.update(week)
+                week_with_author.update(contributor_activity["author"])
+                yield week_with_author
+
+    schema = th.PropertiesList(
+        # Parent keys
+        th.Property("repo", th.StringType),
+        th.Property("org", th.StringType),
+        th.Property("user_id", th.IntegerType),
```
laurentS, line 1347: Do we actually receive `user_id` from the parent, if the parent is a repo?

laurentS (approved):
My comments are mostly suggestions for improvement and discussion.

tap_github/repository_streams.py
laurentS: I renamed `id` to `user_id` in the spirit of clarity as above.

tap_github/repository_streams.py
laurentS: we don't :)

tap_github/repository_streams.py
ericboucher: I don't think that's relevant, we get data for the past year each time, so ultimately we will get the data we want. The only issue is if we ALWAYS get a 202...

tap_github/repository_streams.py
```diff
@@ -1307,3 +1307,60 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+        }
+        parsed_response = super().parse_response(response)
+        for contributor_activity in parsed_response:
+            weekly_data = contributor_activity["weeks"]
+            for week in weekly_data:
+                # no need to save weeks with no contributions.
+                if sum(week[key] for key in ["a", "c", "d"]) == 0:
+                    continue
+                week_with_author = {replacement_keys[k]: v for k, v in week.items()}
+                week_with_author.update(week)
```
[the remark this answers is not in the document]
ericboucher: indeed

PR #72, opened 2022-01-26
Use literal reaction names [breaking]

### LINKED ISSUE #64: Reactions +1 and -1 field names cause downstream issues
I'm trying to load this tap-github into target-bigquery and get the following error

```text
CRITICAL Duplicate field(s) in stream issues:          & reactions.-1 are read as REACTIONS._1 by BigQuery
```

The `pull_requests` stream has the same problem. What's the best way to deal with it? Having the special chars +/- in the schema name causes trouble, should this be fixed at the extractor level or should I look for a solution with this particular loader?

ericboucher:
Following on @laurentS option to fix #64 by using a literal version of reaction emojis +1/-1 to plus_one/minus_one

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -569,6 +569,10 @@ class IssuesStream(GitHubStream):
             # such chars are removed from the data before we pass it on to
             # the target
             row["body"] = row["body"].replace("\x00", "")
+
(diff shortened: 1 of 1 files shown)


laurentS (approved):
LGTM!

PR #74, opened 2022-01-26
Update test logic to avoid running workflows twice

### LINKED ISSUE #73: CI jobs all run twice on PRs
I just noticed after pushing a couple of commits to #71 that each CI job is run twice, once for the PR, once for the push.
There must be some setting on github actions to avoid this, even if we're caching results, it's seems super wasteful.

ericboucher:
Fixes #73

### DIFF
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -1,6 +1,11 @@
 name: Test tap-github
 
-on: [push, pull_request]
+on:
+  # Run on all pull requets and on pushes to master.
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/test_tap.yml
```diff
@@ -1,6 +1,11 @@
 name: Test tap-github
 
-on: [push, pull_request]
+on:
+  # Run on all pull requets and on pushes to master.
```
laurentS, line 4:
```suggestion
  # Run on all pull requests and on pushes to master.
```

laurentS (approved):
Gtg!

edgarrmondragon (approved):

PR #76, opened 2022-02-02
Change order of columns in pkey indexes to improve db performance

laurentS:
We've had a database review for performance bottlenecks by a postgresql expert, and one of the recommendations was to swap the order of index columns in the indexes created on the tables that come out of this tap + `target-postgres`.

The commit in this PR shows how the columns were reordered, allowing for faster updates. The gist of it is that the columns with highest cardinality (most distinct values) should come first, followed by lower cardinality.

This change seems to have helped on our setup.

2 caveats:
- the exact order might vary somewhat depending on what data is being fetched. The order below seems to work for us at this point, but should work in most cases (it seems fairly obvious that there are more users than orgs for instance).
- This might be pretty specific to the way the indexes are created in postgresql, with the specific target we're using (github.com/datamill-co/target-postgres/), and may not apply to other databases/targets. @aaronsteers @edgarrmondragon happy to have your thoughts on whether this is too niche to merge. As a counter argument, it probably won't hurt if it doesn't help :)

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1197,7 +1197,7 @@ class ContributorsStream(GitHubStream):
 
     name = "contributors"
     path = "/repos/{org}/{repo}/contributors"
-    primary_keys = ["org", "repo", "node_id"]
+    primary_keys = ["node_id", "repo", "org"]
     parent_stream_type = RepositoryStream
     ignore_parent_replication_key = False
     state_partitioning_keys = ["repo", "org"]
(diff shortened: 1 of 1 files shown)


aaronsteers:
@laurentS - My first reaction was that it seemed odd to be tuning the target performance by making changes to the target. Then I looked at the text of the PR and I started to come along to this approach.

I think from a design perspective, it is just as reasonable to have a "bottom-up" approach of listing the most granular keys followed by the higher level keys, versus what we have now which is a "top-down" listing of keys in increasing level of specificity. In other words, when given the choice of having `["order-id", "customer-id", "product-id"] as the keys in a sales table, rather than `["product-id", "order-id", "customer-id"]` or similar.

So, long story short, I'm in support of putting the most granular key first, rather than having the highest-level attribute first. I think it can have performance improvements downstream, but also it seems totally reasonable to me otherwise in the abstract.

@edgarrmondragon - Any other/differing thoughts on this topic?

edgarrmondragon:
Changing the order of columns in the composite key shouldn't impact anyone, and if it helps with performance in the target I'm all for it.

@aaronsteers I agree with what all you said above 👍. This might make a good candidate for a "Best Practices SIP", wdyt?

aaronsteers:
Yes, agreed, @edgarrmondragon - thanks!

ericboucher (approved):

PR #78, opened 2022-02-04
Initialize UserContributedToStream using Graphql

### LINKED ISSUE #48: Implement a base GraphQL stream to reduce unnecessary API calls
Currently we have a REST base stream implemented for our tap:
https://github.com/MeltanoLabs/tap-github/blob/main/tap_github/client.py#L10

However, it would be greatly beneficial if we also considered making a base stream for the Github GraphQL API. This is because the Graphql API uses a ['credits' system](https://docs.github.com/en/graphql/overview/resource-limitations) which makes it a lot more adaptive at collecting granular data compared to REST where each resource, no matter how small, takes up an entire request.

Streams that can greatly benefit from the GraphQL API:
- PR Files stream (WIP in #45)
- PRs in general #49

ericboucher:
Initialize UserContributedToStream using Graphql

This also fixes #48


tap_github/user_streams.py
```diff
@@ -158,3 +158,51 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+                    login
+                  }
+                }
+              } 
+            }    
+          }
+        """
+
+    schema = th.PropertiesList(
+        th.Property("id", th.StringType),
```
laurentS, line 197: From what I [understand](https://docs.github.com/en/graphql/guides/using-global-node-ids), `id` in github's graphql context is the same as `nodeId` in REST. Would it be worth renaming the properties here to give a consistent naming in the output of the tap? REST has both `id` (an integer) and `node_id` (string).

laurentS (approved):
Nice! I like that this consumes from a different quota than the REST calls :)

tap_github/client.py
```diff
@@ -188,3 +189,41 @@ def parse_response(self, response: requests.Response) -> Iterable[dict]:
...
+            response: A raw `requests.Response`_ object.
+
+        Yields:
+            One item for every item found in the response.
+
+        .. _requests.Response:
+            https://docs.python-requests.org/en/latest/api/#requests.Response
+        """
+        resp_json = response.json()
+        data = glom(resp_json["data"], self.query_path)
```
edgarrmondragon, line 218: TIL `glom` 🙂 

The SDK already supports [JSONPath](https://sdk.meltano.com/en/latest/dev_guide.html?highlight=jsonpath#restful-jsonpaths), so I'm curious what are the benefits of glom over that.

edgarrmondragon (changes requested):

tap_github/user_streams.py
```diff
@@ -158,3 +158,51 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
         ),
     ).to_dict()
+
+
+class UserContributedToStream(GitHubGraphqlStream):
+    """Defines 'UserContributedToStream' stream. Warning: this stream 'only' gets the first 100 projects (by stars)."""
+
+    name = "user_contributed_to"
+    query_path = "user.repositoriesContributedTo.nodes"
+    primary_keys = ["username"]
```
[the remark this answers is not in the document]
ericboucher: sadly, I did not find any...

tap_github/client.py
ericboucher: @edgarrmondragon I am trying to use JSONPath and it works with online tools but I can't get it to work in the code.

This is what I am doing:

resp_json = response.json()
yield from extract_jsonpath(self.query_jsonpath, input=resp_json)

with `query_jsonpath = "$.data.user.repositoriesContributedTo.nodes.*"`

ericboucher: nvm I think I fixed it in https://github.com/MeltanoLabs/tap-github/pull/78/commits/9435df7addb65237d375f543ba83f626a214456c

edgarrmondragon: @ericboucher looks good!

edgarrmondragon (approved):

PR #79, opened 2022-02-04
chore: update meltano.yml

edgarrmondragon:
Update meltano.yml

### DIFF
--- a/meltano.yml
+++ b/meltano.yml
@@ -1,26 +1,43 @@
 version: 1
-send_anonymous_usage_stats: true
+send_anonymous_usage_stats: false
 project_id: 96584f7b-a36c-46e0-b41a-7f9074293137
 plugins:
   extractors:
   - name: tap-github
     namespace: tap_github
-    pip_url: blah
-    executable: /Users/ajsteers/Source/tap-github/tap-github.sh
+    pip_url: -e .
     capabilities:
     - state
     - catalog
     - discover
     settings:
-    - name: searches
-      kind: array
+    - name: user_agent
+      kind: string
+    - name: metrics_log_level
+      kind: string
     - name: auth_token
       kind: password
+    - name: additional_auth_tokens
+      kind: array
+    - name: rate_limit_buffer
+      kind: array
+    - name: searches
+      kind: array
+    - name: organizations
+      kind: array
+    - name: repositories
+      kind: array
+    - name: user_usernames
+      kind: array
+    - name: user_ids
+      kind: array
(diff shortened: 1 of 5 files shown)


tap_github/tap.py
```diff
@@ -49,7 +46,7 @@ class TapGitHub(Tap):
         ),
         th.Property(
             "rate_limit_buffer",
-            th.ArrayType(th.IntegerType),
+            th.IntegerType,
```
edgarrmondragon, line 49: @ericboucher seeing how this is used I think it's safe to assume it's an integer and not an array

pyproject.toml
```diff
@@ -38,7 +39,11 @@ multi_line_output = 3 # Vertical Hanging Indent
 src_paths = "tap_github"
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+# Uncomment the pinned version in favor of the git URL once
+# https://github.com/python-poetry/poetry-core/pull/257 is merged
+# and a new poetry-core 1.0.x is released
+# requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core @ git+https://github.com/python-poetry/poetry-core.git@master"]
```
edgarrmondragon, line 46: This lets us set `pip_url: -e .` in `meltano.yml` for editable installation 🙂

meltano.yml
```diff
@@ -1,26 +1,43 @@
...
+    - name: user_agent
+      kind: string
+    - name: metrics_log_level
+      kind: string
     - name: auth_token
       kind: password
+    - name: additional_auth_tokens
+      kind: array
+    - name: rate_limit_buffer
+      kind: array
```
ericboucher, line 23: should this also be updated @edgarrmondragon ?

ericboucher (commented):
Minor type change in meltano.yml, otherwise gtg

ericboucher (approved):

pyproject.toml
laurentS: This scares me, as you're not really pinning your build system version at all here. Is it on purpose?

meltano.yml
edgarrmondragon: it should 😄

pyproject.toml
edgarrmondragon: @laurentS  Me too 😅. I don't think anyone uses this meltano file is used in production, though so I'd be ok with referencing a branch and watching closely for a pypi release of poetry-core with pep 660 support. wdyt?

PR #80, opened 2022-02-04
Silence unnecessary replication_key warnings

ericboucher:
Tweaking our warning for replication_key handling since "commit_timestamp" and "created_at" work by default.

And leveraging our custom "since" logic to exit early with the "created_at" replication_key


laurentS (approved):
LGTM!

PR #84, opened 2022-02-08
Add parent replication_key

ericboucher:
After merging #80, I noticed a ?weird? behavior with replication keys that I think stems from the ignore_parent_replication_key and `_set_compatible_replication_methods` [sdk](https://gitlab.com/meltano/sdk/-/blob/main/singer_sdk/tap_base.py#L330)

This PR is an attempt to fix a warning that pops up but doesn't feel normal:

`Beginning incremental sync of 'pull_requests' with context: {'org': 'ourresearch', 'repo': 'oadoi'}...`
`The replication key 'None' is not fully supported by this client yet.` which happens to almost all our streams.

@aaronsteers you're probably the most versed in replication_keys, so very curious to get your thoughts.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -15,6 +15,9 @@ class RepositoryStream(GitHubRestStream):
     MAX_RESULTS_LIMIT = 1000
 
     name = "repositories"
+    # updated_at will be updated any time the repository object is updated,
+    # e.g. when the description or the primary language of the repository is updated.
+    replication_key = "updated_at"
 
     def get_url_params(
         self, context: Optional[dict], next_page_token: Optional[Any]
@@ -171,7 +174,7 @@ class ReadmeStream(GitHubRestStream):
     path = "/repos/{org}/{repo}/readme"
     primary_keys = ["repo", "org"]
     parent_stream_type = RepositoryStream
-    ignore_parent_replication_key = False
+    ignore_parent_replication_key = True
     state_partitioning_keys = ["repo", "org"]
     tolerated_http_errors = [404]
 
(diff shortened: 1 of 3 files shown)


aaronsteers:
@ericboucher - It's been a while since I've thought hard on this topic and the edge cases. As you may already know, the `ignore_parent_replication_key` setting allows the developer to specify if looping through new parent records is sufficient to also catch new child records. When `ignore_parent_replication_key=True`, the inference is that we can't tell if there are new child records simply by looking for updated parent records.

Can you help me by confirming at least one example of each for our discussion?

- `EventsStream`
   - `ignore_parent_replication_key = True`
   - Parent is `RepositoryStream`
   - Description: Events can be added without bumping the updated timestamp on the Repository.
- `LanguagesStream`
   - `ignore_parent_replication_key = False`
   - Parent is `RepositoryStream`
   - Description: Code languages won't update without _also_ bumping the updated timestamp on the Repository.

Do those two descriptions look correct to you? I may need to dig into the code a bit but it's always helpful to start with a couple real-life scenarios.

Might be a tangent, but just for completeness: if memory serves, this is more complex when the replication key is _not_ a timestamp, but is some other column type, such as an integer. (Related to how "now" works and what we can infer from datetime column's relationship between parent and child records that might have different column names.)

ericboucher:
@aaronsteers that looks correct for Events and Language stream based on my research. Although GitHub's doc in that regard is clearly lacking...

All our replication_keys are timestamps for now

tap_github/repository_streams.py
```diff
@@ -15,6 +15,9 @@ class RepositoryStream(GitHubRestStream):
     MAX_RESULTS_LIMIT = 1000
 
     name = "repositories"
+    # updated_at will be updated any time the repository object is updated,
+    # e.g. when the description or the primary language of the repository is updated.
+    replication_key = "updated_at"
```
ericboucher, line 20: The other option would be to used "pushed_at" instead

ericboucher: or not use any replication_key for repo and set all ignore_parent_replication_key to True for sub streams?

laurentS (approved):

PR #85, opened 2022-02-08
Fix prop names and missing pkey property on UserContributedToStream

laurentS:
This PR corrects some small errors that slipped through code review:
- property names are snake_cased
- `username` was missing in properties

### DIFF
--- a/tap_github/user_streams.py
+++ b/tap_github/user_streams.py
@@ -181,24 +181,25 @@ class UserContributedToStream(GitHubGraphqlStream):
               repositoriesContributedTo (first: 100 includeUserRepositories: true orderBy: {field: STARGAZERS, direction: DESC}) {
                 nodes {
                   node_id: id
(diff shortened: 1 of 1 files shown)


tap_github/user_streams.py
```diff
@@ -181,24 +181,25 @@ def query(self) -> str:
               repositoriesContributedTo (first: 100 includeUserRepositories: true orderBy: {field: STARGAZERS, direction: DESC}) {
                 nodes {
                   node_id: id
-                  nameWithOwner
-                  openGraphImageUrl
-                  stargazerCount
+                  name_with_owner: nameWithOwner
+                  open_graph_image_url: openGraphImageUrl
```
ericboucher, line 185: I think this has some other name in rest

laurentS: I'll leave it as is for now.

PR #90, opened 2022-02-25
Retry on 401 unauthorized errors

ericboucher:
Unfortunately, the GitHub API returns 401 unauthorized errors from time to time. Randomly...

To mitigate that, one option is to make 401 a retriable error for now.


laurentS (approved):
It would be good to understand why these errors are happening, but it seems that's an issue for github to solve :)

PR #91, opened 2022-02-25
Add config option to skip API calls to repositories

laurentS:
This PR tries to address https://gitlab.com/meltano/sdk/-/issues/217 within `tap-github` using a specific solution.
It might be possible to generalise the idea and add something equivalent in the sdk.

This has become fairly high priority for us:
- we sync different streams independently (1 by 1) at different frequencies (some hourly, some daily, some less often). Currently all child streams force a call to the `/repo` endpoint.
- github API quota usage is essentially doubled in incremental mode. eg: syncing `issues` for 1 repo yields 2 API calls, 1 to repos, 1 to issues, and the result of the first call is discarded, which seems wasteful.
- we have seen an increase in random 401 errors from the API (despite valid auth, and remaining within rate limits), these lead to the tap crashing with no real solution we can implement (see https://gitlab.com/meltano/sdk/-/issues/282). While this PR does not directly fix this, it should help by reducing the number of API calls and potential for errors.
- it's just nice to avoid doing work that we don't need :)

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -77,15 +77,15 @@ class RepositoryStream(GitHubRestStream):
     def get_records(self, context: Optional[dict]) -> Iterable[Dict[str, Any]]:
         """
         Override the parent method to allow skipping API calls
-        if the stream is deselected and skip_repo_for_child_streams is True in config.
+        if the stream is deselected and skip_parent_streams is True in config.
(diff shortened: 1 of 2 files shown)


ericboucher:
@laurentS how about doing something slightly more general already such as `skip_parent_streams`. We will likely be facing the same issue for user child streams

ericboucher (approved):

Issue #89, opened 2022-02-09
Add pagination for the Graphql client


ericboucher:
The basic implementation done in #78 does not allow for pagination. If we are to use the graphql endpoint to get more data, we will need to implement pagination in the client.

ericboucher:
https://stackoverflow.com/questions/48116781/github-api-v4-how-can-i-traverse-with-pagination-graphql

ericboucher:
The pagination should be implemented in a way that allows for "deep pagination". I.e. look for the deepest next page token and rotate it until `hasNextPage = false`

PR #93, opened 2022-03-10
Add missing endpoints available in singer-io/tap-github

### LINKED ISSUE #83: Add all endpoints available in singer-io/tap-github
Effectively, this tap is already a lot more advanced than https://github.com/singer-io/tap-github, which is the tap currently proposed as a default GitHub tap on meltano.com.

Let's add missing endpoints from https://github.com/singer-io/tap-github and reach out to the relevant persons at Meltano to upgrade the default tap to this one.

Missing endpoints:
- Assignees
- Collaborators
- Reviews
- Review Comments
- ...

Ry-DS:
### Closes #83

TODO:
- [ ] Review (Meltano)
- [x] Discuss the code duplication (CI SonarCloud)
- [x] ~~Convert Reviews and Pull Request Commits to GraphQL streams (they are too slow with rest since PR is the parent)~~ See https://github.com/MeltanoLabs/tap-github/pull/93#issuecomment-1072421606
- [x] ~~#89~~ Decided that pagination of the graphql api is not worthwhile to explore (see above comment) 
- [x] Test all streams individually to make sure they give some output
- [x] (Maybe) write test just for organization
- [x] Make sure the following are correct
    - [x] ~~Parent keys are missing in the new streams. Is this ok?~~ Decided to add them
    - [x] Set `ignore_parent_replication_key` to true unless sure if not
- [x] Finish implementing all streams

Streams:
- [x] Collaborators
- [x] Assignees
- [x] Reviews
- [x] Review Comments
- [x] Commit Comments
- [x] Issue Milestones
- [x] Projects 
- [x] Project Cards
- [x] Project Columns 
- [x] PR Commits
- [x] Releases
- [x] Organization stream
    - [x] Teams
    - [x] Team members
    - [x] Team memberships

This PR involved a lot of json to propertylist conversions. Here are some regex conversion expressions to make things easier:
- Automatic string type conversion: `"(\w+)": .+"` `th.Property("$1", th.StringType)`
- Integer type conversion: `"(\w+)": \d+` `th.Property("$1", th.IntegerType)`
- Boolean type conversion: `"(\w+)": (true|false)+` `th.Property("$1", th.BooleanType)`
- Object type conversion `"(\w+)": \{(.|\n+?)\},` `th.Property("$1", th.ObjectType($2),),`

Maybe we could standardise the above regex queries and document them in the README?

Discussion points:
- We could use the graphql api for Pull Request Reviews if we want to extend our quota potential. Thoughts? Went with the path of least resistance for now (REST) but if you do see huge savings by using graphql let me know.
- Not sure how to fix the test as it now needs authentication to work (collaborators needs a github token to work)
    - https://github.com/MeltanoLabs/tap-github/runs/5497860148?check_suite_focus=true#step:10:17131
    - `singer_sdk.exceptions.FatalAPIError: 403 Client Error: Forbidden for path: /repos/octocat/Hello-World/collaborators`


ericboucher:
@Ry-DS make sure to have a look @edgarrmondragon's PR https://github.com/MeltanoLabs/tap-github/pull/31 to avoid duplication

Ry-DS:
Conversions using regex patterns in description (some need manual fixes)
<details>
   <summary>Click to expand</summary>

   ```python
// projects
{
    th.Property("owner_url", th.StringType),
    th.Property("url", th.StringType),
    th.Property("html_url", th.StringType),
    th.Property("columns_url", th.StringType),
    th.Property("id", th.IntegerType),
    th.Property("node_id", th.StringType),
    th.Property("name", th.StringType),
    th.Property("body", th.StringType),
    th.Property("number", th.IntegerType),
    th.Property("state", th.StringType),
    th.Property("creator", th.ObjectType(
      th.Property("login", th.StringType),
      th.Property("id", th.IntegerType),
      th.Property("node_id", th.StringType),
      th.Property("avatar_url", th.StringType),
      th.Property("gravatar_id", th.StringType),
      th.Property("url", th.StringType),
      th.Property("html_url", th.StringType),
      th.Property("followers_url", th.StringType),
      th.Property("following_url", th.StringType),
      th.Property("gists_url", th.StringType),
      th.Property("starred_url", th.StringType),
      th.Property("subscriptions_url", th.StringType),
      th.Property("organizations_url", th.StringType),
      th.Property("repos_url", th.StringType),
      th.Property("events_url", th.StringType),
      th.Property("received_events_url", th.StringType),
      th.Property("type", th.StringType),
      th.Property("site_admin", th.BooleanType)
    ),),
    th.Property("created_at", th.StringType),
    th.Property("updated_at", th.StringType)
  }

// project cards
{
    th.Property("url", th.StringType),
    th.Property("id", th.IntegerType),
    th.Property("node_id", th.StringType),
    th.Property("note", th.StringType),
    th.Property("creator", th.ObjectType(
      th.Property("login", th.StringType),
      th.Property("id", th.IntegerType),
      th.Property("node_id", th.StringType),
      th.Property("avatar_url", th.StringType),
      th.Property("gravatar_id", th.StringType),
      th.Property("url", th.StringType),
      th.Property("html_url", th.StringType),
      th.Property("followers_url", th.StringType),
      th.Property("following_url", th.StringType),
      th.Property("gists_url", th.StringType),
      th.Property("starred_url", th.StringType),
      th.Property("subscriptions_url", th.StringType),
      th.Property("organizations_url", th.StringType),
      th.Property("repos_url", th.StringType),
      th.Property("events_url", th.StringType),
      th.Property("received_events_url", th.StringType),
      th.Property("type", th.StringType),
      th.Property("site_admin", th.BooleanType)
    ),),
    th.Property("created_at", th.StringType),
    th.Property("updated_at", th.StringType),
    th.Property("archived", th.BooleanType),
    th.Property("column_url", th.StringType),
    th.Property("content_url", th.StringType),
    th.Property("project_url", th.StringType)
  }
// project columns
{
    th.Property("url", th.StringType),
    th.Property("project_url", th.StringType),
    th.Property("cards_url", th.StringType),
    th.Property("id", th.IntegerType),
    th.Property("node_id", th.StringType),
    th.Property("name", th.StringType),
    th.Property("created_at", th.StringType),
    th.Property("updated_at", th.StringType)
  }
// pr commits
{
    th.Property("url", th.StringType),
    th.Property("sha", th.StringType),
    th.Property("node_id", th.StringType),
    th.Property("html_url", th.StringType),
    th.Property("comments_url", th.StringType),
    th.Property("commit", th.ObjectType(
      th.Property("url", th.StringType),
      th.Property("author", th.ObjectType(
        th.Property("name", th.StringType),
        th.Property("email", th.StringType),
        th.Property("date", th.StringType)
      ),),
      "committer": {
        th.Property("name", th.StringType),
        th.Property("email", th.StringType),
        th.Property("date", th.StringType)
      ),),
      th.Property("message", th.StringType),
      th.Property("tree", th.ObjectType(
        th.Property("url", th.StringType),
        th.Property("sha", th.StringType)
      ),),
      th.Property("comment_count", th.IntegerType),
      th.Property("verification", th.ObjectType(
        th.Property("verified", th.BooleanType),
        th.Property("reason", th.StringType),
        "signature": null,
        "payload": null
      }
    ),),
    th.Property("author", th.ObjectType(
      th.Property("login", th.StringType),
      th.Property("id", th.IntegerType),
      th.Property("node_id", th.StringType),
      th.Property("avatar_url", th.StringType),
      th.Property("gravatar_id", th.StringType),
      th.Property("url", th.StringType),
      th.Property("html_url", th.StringType),
      th.Property("followers_url", th.StringType),
      th.Property("following_url", th.StringType),
      th.Property("gists_url", th.StringType),
      th.Property("starred_url", th.StringType),
      th.Property("subscriptions_url", th.StringType),
      th.Property("organizations_url", th.StringType),
      th.Property("repos_url", th.StringType),
      th.Property("events_url", th.StringType),
      th.Property("received_events_url", th.StringType),
      th.Property("type", th.StringType),
      th.Property("site_admin", th.BooleanType)
    ),),
    th.Property("committer", th.ObjectType(
      th.Property("login", th.StringType),
      th.Property("id", th.IntegerType),
      th.Property("node_id", th.StringType),
      th.Property("avatar_url", th.StringType),
      th.Property("gravatar_id", th.StringType),
      th.Property("url", th.StringType),
      th.Property("html_url", th.StringType),
      th.Property("followers_url", th.StringType),
      th.Property("following_url", th.StringType),
      th.Property("gists_url", th.StringType),
      th.Property("starred_url", th.StringType),
      th.Property("subscriptions_url", th.StringType),
      th.Property("organizations_url", th.StringType),
      th.Property("repos_url", th.StringType),
      th.Property("events_url", th.StringType),
      th.Property("received_events_url", th.StringType),
      th.Property("type", th.StringType),
      th.Property("site_admin", th.BooleanType)
    ),),
    "parents": [
      {
        th.Property("url", th.StringType),
        th.Property("sha", th.StringType)
      }
    ]
  }


// releases
{
    th.Property("url", th.StringType),
    th.Property("html_url", th.StringType),
    th.Property("assets_url", th.StringType),
    th.Property("upload_url", th.StringType),
    th.Property("tarball_url", th.StringType),
    th.Property("zipball_url", th.StringType),
    th.Property("id", th.IntegerType),
    th.Property("node_id", th.StringType),
    th.Property("tag_name", th.StringType),
    th.Property("target_commitish", th.StringType),
    th.Property("name", th.StringType),
    th.Property("body", th.StringType),
    th.Property("draft", th.BooleanType),
    th.Property("prerelease", th.BooleanType),
    th.Property("created_at", th.StringType),
    th.Property("published_at", th.StringType),
    th.Property("author", th.ObjectType(
      th.Property("login", th.StringType),
      th.Property("id", th.IntegerType),
      th.Property("node_id", th.StringType),
      th.Property("avatar_url", th.StringType),
      th.Property("gravatar_id", th.StringType),
      th.Property("url", th.StringType),
      th.Property("html_url", th.StringType),
      th.Property("followers_url", th.StringType),
      th.Property("following_url", th.StringType),
      th.Property("gists_url", th.StringType),
      th.Property("starred_url", th.StringType),
      th.Property("subscriptions_url", th.StringType),
      th.Property("organizations_url", th.StringType),
      th.Property("repos_url", th.StringType),
      th.Property("events_url", th.StringType),
      th.Property("received_events_url", th.StringType),
      th.Property("type", th.StringType),
      th.Property("site_admin", th.BooleanType)
    ),),
    "assets": [
      {
        th.Property("url", th.StringType),
        th.Property("browser_download_url", th.StringType),
        th.Property("id", th.IntegerType),
        th.Property("node_id", th.StringType),
        th.Property("name", th.StringType),
        th.Property("label", th.StringType),
        th.Property("state", th.StringType),
        th.Property("content_type", th.StringType),
        th.Property("size", th.IntegerType),
        th.Property("download_count", th.IntegerType),
        th.Property("created_at", th.StringType),
        th.Property("updated_at", th.StringType),
        "uploader": {
          th.Property("login", th.StringType),
          th.Property("id", th.IntegerType),
          th.Property("node_id", th.StringType),
          th.Property("avatar_url", th.StringType),
          th.Property("gravatar_id", th.StringType),
          th.Property("url", th.StringType),
          th.Property("html_url", th.StringType),
          th.Property("followers_url", th.StringType),
          th.Property("following_url", th.StringType),
          th.Property("gists_url", th.StringType),
          th.Property("starred_url", th.StringType),
          th.Property("subscriptions_url", th.StringType),
          th.Property("organizations_url", th.StringType),
          th.Property("repos_url", th.StringType),
          th.Property("events_url", th.StringType),
          th.Property("received_events_url", th.StringType),
          th.Property("type", th.StringType),
          th.Property("site_admin", th.BooleanType)
        }
      }
    ]
  }
// teams
 {
    th.Property("id", th.IntegerType),
    th.Property("node_id", th.StringType),
    th.Property("url", th.StringType),
    th.Property("html_url", th.StringType),
    th.Property("name", th.StringType),
    th.Property("slug", th.StringType),
    th.Property("description", th.StringType),
    th.Property("privacy", th.StringType),
    th.Property("permission", th.StringType),
    th.Property("members_url", th.StringType),
    th.Property("repositories_url", th.StringType),
    "parent": null
  }

// team members
{
    th.Property("login", th.StringType),
    th.Property("id", th.IntegerType),
    th.Property("node_id", th.StringType),
    th.Property("avatar_url", th.StringType),
    th.Property("gravatar_id", th.StringType),
    th.Property("url", th.StringType),
    th.Property("html_url", th.StringType),
    th.Property("followers_url", th.StringType),
    th.Property("following_url", th.StringType),
    th.Property("gists_url", th.StringType),
    th.Property("starred_url", th.StringType),
    th.Property("subscriptions_url", th.StringType),
    th.Property("organizations_url", th.StringType),
    th.Property("repos_url", th.StringType),
    th.Property("events_url", th.StringType),
    th.Property("received_events_url", th.StringType),
    th.Property("type", th.StringType),
    th.Property("site_admin", th.BooleanType)
  }

// team roles
{
  th.Property("url", th.StringType),
  th.Property("role", th.StringType),
  th.Property("state", th.StringType)
}
   ```
</details>

tap_github/repository_streams.py
```diff
@@ -1407,3 +1919,125 @@ def parse_response(self, response: requests.Response) -> Iterable[dict]:
         th.Property("type", th.StringType),
         th.Property("site_admin", th.BooleanType),
     ).to_dict()
+
+
+class ProjectsStream(GitHubRestStream):
+    name = "projects"
+    path = "/repos/{org}/{repo}/projects"
+    # TODO I think this can be true, but I'm not sure.
```
ericboucher, line 1927: It's probably True most of the time. Very few children updates actually change the "updated_at" field of the parent (repository)

todo_streams.py
```diff
@@ -0,0 +1,60 @@
+from singer_sdk import typing as th
+
+from tap_github.client import GitHubRestStream
+
+
+# TODO not sure what the parent stream is here.
```
ericboucher, line 6: organization?

Ry-DS:
Can't get CI to pass because of the following error:
`  singer_sdk.exceptions.FatalAPIError: 403 Client Error: Forbidden for path: /orgs/MeltanoLabs/teams`
any ideas?

ericboucher:
@Ry-DS does it work locally? It probably requires an authentication with access to org level data. I think we should either:
1. bypass this test for now or put it inside a try catch that outputs a warning instead of a failure - but test it thoroughly locally
2. add a GitHub token with access to read the Meltano org data and use it specifically for this test. Eg. by overriding the "GITHUB_TOKEN" env temporarily. @aaronsteers what do you think?

tap_github/tests/test_tap.py
```diff
@@ -1,25 +1,17 @@
...
-    deselect_all_streams,
-    set_catalog_stream_selected,
-)
 
 from tap_github.tap import TapGitHub
 
 from .fixtures import repo_list_config
 
-repo_list_2 = ["octocat/hello-world", "MeltanoLabs/tap-github", "mapswipe/mapswipe"]
+repo_list_2 = ["MeltanoLabs/tap-github"]
```
ericboucher, line 7: Why the switch to one project?

Ry-DS: the new collaborators stream needs authentication. Since CI only has access to MeltanoLabs, we can't test the other projects easily anymore as CI doesn't have access to them. What do you reckon we do?

Ry-DS:
> @Ry-DS does it work locally? It probably requires an authentication with access to org level data. I think we should either:
> 
> 1. bypass this test for now or put it inside a try catch that outputs a warning instead of a failure - but test it thoroughly locally
> 2. add a GitHub token with access to read the Meltano org data and use it specifically for this test. Eg. by overriding the "GITHUB_TOKEN" env temporarily. @aaronsteers what do you think?

Tested locally and works fine. Option 2. would be the best. Not too sure either way on how to block out testing just the teams route as the tests are generated as a function array:
https://github.com/MeltanoLabs/tap-github/blob/47dc14d8edf311381931970633d8418012f06af1/tap_github/tests/test_core.py#L41

Ry-DS:
I ran the events stream on a whim and realised the id fields were mislabelled. I reckon it might be worthwhile to actually try load the tap data into a db with a simple target such as `target-sqlite` to ensure all rows are correctly typed, as there is nothing right now to confirm this.

ericboucher:
@Ry-DS It looks like the API is indeed returning them as strings, but they are clearly integers... Can we force the conversion?
Do you get an error?

-> https://api.github.com/repos/facebook/react/events

Ry-DS:
> @Ry-DS It looks like the API is indeed returning them as strings, but they are clearly integers... Can we force the conversion? Do you get an error?
> 
> -> https://api.github.com/repos/facebook/react/events

It makes sense that it's a string as the number is quite large (20 billion) which definitely won't fit in a 32bit signed integer (max 2 billion). I think it's the right move to keep it as strings.

tap_github/repository_streams.py
```diff
@@ -404,14 +404,14 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
         th.Property(
             "target_repo",
             th.ObjectType(
-                th.Property("id", th.StringType),
+                th.Property("id", th.IntegerType),
```
ericboucher, line 407: It seems odd that the parent repo id would be a string and the target repo id an integer?

tap_github/repository_streams.py
```diff
@@ -1222,6 +1525,252 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+        ),
+    ).to_dict()
+
+
+class ReviewsStream(GitHubRestStream):
+    name = "reviews"
+    path = "/repos/{org}/{repo}/pulls/{pull_number}/reviews"
+    primary_keys = ["id"]
+    parent_stream_type = PullRequestsStream
+    ignore_parent_replication_key = True
```
ericboucher, line 1649: It might be worth checking if a review changes the "updated_at" field for a PR

tap_github/organization_streams.py
```diff
@@ -0,0 +1,168 @@
...
+    def path(self) -> str:  # type: ignore
+        """Return the API endpoint path."""
+        return "/orgs/{org}"
+
+    @property
+    def partitions(self) -> Optional[List[Dict]]:
+        """Return a list of partitions."""
+        return [{"org": org} for org in self.config["organizations"]]
+
+    def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
```
laurentS, line 23: I think `dict` as a type hint only works for python 3.9+. Before that you need `Dict` (same story for `List`, etc...)(https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#built-in-types).

tap_github/organization_streams.py
```diff
@@ -0,0 +1,168 @@
...
+from singer_sdk import typing as th  # JSON Schema typing helpers
+
+from tap_github.client import GitHubRestStream
+
+
+class OrganizationStream(GitHubRestStream):
+    name = "organizations"
+
+    @property
+    def path(self) -> str:  # type: ignore
```
laurentS, line 14: Is there a reason this one is a property as opposed to a simple string like in other streams?

tap_github/organization_streams.py
```diff
@@ -0,0 +1,168 @@
...
+    ).to_dict()
+
+
+class TeamMembersStream(GitHubRestStream):
+    name = "team_members"
+    primary_keys = ["id"]
+    path = "/orgs/{org}/teams/{team_slug}/members"
+    ignore_parent_replication_key = True
+    parent_stream_type = TeamsStream
+    state_partitioning_keys = ["org", "team_slug"]
```
laurentS, line 114: With [this PR](https://github.com/MeltanoLabs/tap-github/pull/76) we've done the route of ordering keys by decreasing expected cardinality.
```suggestion
    state_partitioning_keys = ["team_slug", "org"]
```
> columns with highest cardinality (most distinct values) should come first, followed by lower cardinality.

tap_github/organization_streams.py
```diff
@@ -0,0 +1,168 @@
...
+    ).to_dict()
+
+
+class TeamRolesStream(GitHubRestStream):
+    name = "team_roles"
+    path = "/orgs/{org}/teams/{team_slug}/memberships/{username}"
+    ignore_parent_replication_key = True
+    primary_keys = ["url"]
+    parent_stream_type = TeamMembersStream
+    state_partitioning_keys = ["org", "team_slug", "username"]
```
laurentS, line 157:
```suggestion
    state_partitioning_keys = ["username", "team_slug", "org"]
```

tap_github/repository_streams.py
```diff
@@ -519,6 +519,159 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+                th.Property("html_url", th.StringType),
+                th.Property("followers_url", th.StringType),
+                th.Property("following_url", th.StringType),
+                th.Property("gists_url", th.StringType),
+                th.Property("starred_url", th.StringType),
+                th.Property("subscriptions_url", th.StringType),
+                th.Property("organizations_url", th.StringType),
+                th.Property("repos_url", th.StringType),
+                th.Property("events_url", th.StringType),
+                th.Property("received_events_url", th.StringType),
```
laurentS, line 563: I'd be of the opinion that all these fields can easily be guessed from username alone, and so may not be worth storing (they are long strings). For all other streams, we've kept these to the non-guessable part, factored out in #98 for easy reuse.

tap_github/repository_streams.py
```diff
@@ -547,6 +700,87 @@ def parse_response(self, response: requests.Response) -> Iterable[dict]:
...
+        th.Property("login", th.StringType),
+        th.Property("id", th.IntegerType),
+        th.Property("node_id", th.StringType),
+        th.Property("avatar_url", th.StringType),
+        th.Property("gravatar_id", th.StringType),
+        th.Property("url", th.StringType),
+        th.Property("html_url", th.StringType),
+        th.Property("followers_url", th.DateTimeType),
+        th.Property("following_url", th.DateTimeType),
+        th.Property("gists_url", th.DateTimeType),
```
laurentS, line 725: Types don't look right here

tap_github/repository_streams.py
```diff
@@ -547,6 +700,87 @@ def parse_response(self, response: requests.Response) -> Iterable[dict]:
...
```
laurentS, line 772: Incorrect types?

tap_github/repository_streams.py
```diff
@@ -1222,6 +1525,252 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
     ).to_dict()
 
 
+class PullRequestCommits(GitHubRestStream):
+    name = "pull_request_commits"
+    path = "/repos/{org}/{repo}/pulls/{pull_number}/commits"
+    ignore_parent_replication_key = True
+    primary_keys = ["node_id"]
+    parent_stream_type = PullRequestsStream
+    state_partitioning_keys = ["pull_number"]
```
laurentS, line 1534:
```suggestion
    state_partitioning_keys = ["org", "repo", "pull_number"]
```
You need all 3, or the state will be shared among all PRs with the same number :(

laurentS: @aaronsteers I've been wondering about this for `tap-gitlab` as well: with cases like this, the state object will become embarrassingly big, with one entry per PR essentially. In our case, we track all PRs over several hundred repos, this state object could end up having well into the ten of thousands of key/values if not more. As the tap emits one state object every N records, this ends up being very verbose (also thinking of memory consumption issues (see [this PR](https://github.com/oviohub/target-postgres/pull/2) for instance)).
Is there any best practice around managing such situations? I'm thinking there might be a way to only limit the state to "one level up", ie: `org, repo` only?

tap_github/repository_streams.py
```diff
@@ -1222,6 +1525,252 @@ def post_process(self, row: dict, context: Optional[dict] = None) -> dict:
...
+    ).to_dict()
+
+
+class ReviewsStream(GitHubRestStream):
+    name = "reviews"
+    path = "/repos/{org}/{repo}/pulls/{pull_number}/reviews"
+    primary_keys = ["id"]
+    parent_stream_type = PullRequestsStream
+    ignore_parent_replication_key = True
+    state_partitioning_keys = ["pull_number"]
```
laurentS, line 1650: See comments above.

tap_github/streams.py
```diff
@@ -0,0 +1,102 @@
+from enum import Enum
+from typing import Type, Set, List
+
+from singer_sdk.streams.core import Stream
+
+from tap_github.organization_streams import OrganizationStream
```
laurentS, line 6: I think this one is duplicated in this file, no?

tap_github/streams.py
```diff
@@ -0,0 +1,102 @@
...
+)
+from tap_github.organization_streams import (
+    OrganizationStream,
+    TeamsStream,
+    TeamMembersStream,
+    TeamRolesStream,
+)
+
+
+class Streams(Enum):
```
laurentS, line 48: I'm confused by why this derives from `Enum`.

tap_github/tests/test_tap.py
laurentS: We're having the same issue with `tap-gitlab`, but no good solution has come up yet.

laurentS (changes requested):
That was a lot! I think there's not much missing, but some changes are probably required for the tap to work.
I've only written some of the comments once to avoid repeating the same thing over and over, but they might apply to multiple streams.

.github/workflows/test_tap.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: ["3.10", 3.9, 3.8, 3.7]
```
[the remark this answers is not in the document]
Ry-DS: I realised the lower versions are more likely to crash, so I reordered them to run from least likely to most likely to crash (so you can recognise if your code works on later versions earlier :eyes:)

tap_github/streams.py
Ry-DS: Without `Enum` I wouldn't get access to features like defining methods to the enum values. E.g: `Streams.ORGANIZATIONS.all_valid_queries()`

tap_github/streams.py
```diff
@@ -0,0 +1,102 @@
...
+    valid_queries: Set[str]
+    streams: List[Type[Stream]]
+
+    def __init__(self, valid_queries: Set[str], streams: List[Type[Stream]]):
+        self.valid_queries = valid_queries
+        self.streams = streams
+
+    REPOSITORY = (
```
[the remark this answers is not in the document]
Ry-DS: not too sure what you mean by this. Do you mean repository_streams.py? If you mean splitting up this enum into several places we might lose functionality such as the `all_valid_queries` which reads all possible streams at once

tap_github/repository_streams.py
Ry-DS: Tried adding comments and such and the PR `updated_at` changes :eyes:

tap_github/organization_streams.py
Ry-DS: good catch, I just copied another stream haha

aaronsteers:
> > @Ry-DS It looks like the API is indeed returning them as strings, but they are clearly integers... Can we force the conversion? Do you get an error?
> > -> https://api.github.com/repos/facebook/react/events
> 
> It makes sense that it's a string as the number is quite large (20 billion) which definitely won't fit in a 32bit signed integer (max 2 billion). I think it's the right move to keep it as strings (it would error out otherwise).

I've logged an issue in the SDK to keep this discussion going. For now, I agree it sounds like preserving as a string type is the "safe" option.

tap_github/tests/fixtures.py
```diff
@@ -37,38 +37,55 @@ def repo_list_config(request):
...
+        "user_ids": user_id_list,
+    }
+
+
+@pytest.fixture
+def organization_list_config(request):
+    """
+    Get a default list of usernames or pass your own by decorating your test with
+    @pytest.mark.user_id_list(['ericboucher', 'aaronsteers'])
+    """
```
aaronsteers, line 82: Was this docstring supposed to be updated to "organizations"? 

Something this this, I think:

```suggestion
def organization_list_config(request):
    """
    Get a default list of organizations or pass your own by decorating your test with
    @pytest.mark.organization_list(['MeltanoLabs', 'SomeOrg'])
    """
```

tap_github/tests/test_tap.py
aaronsteers: @laurentS and @Ry-DS 

Couple questions related to auth and CI tokens:

1. With the current CI token, can we have plural projects here if all are from MeltanoLabs (for instance, adding tap-gitlab and one or two others)?
2. With the current CI token, can we get data from the `MeltanoLabs` organization when pulling the org-related streams?

If "no" to either of these, can we get granular enough control using personal access tokens, or do we need to consider a new "bot" user with the proper grants to make the above two work? Or perhaps there's another path forward?

.github/workflows/test_tap.yml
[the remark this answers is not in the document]
Ry-DS: yup up to you if you want to disable it. Not too sure how github credits work for this project but if it's plentiful, I reckon it's a good idea to disable it.

tap_github/tests/test_tap.py
Ry-DS: Checked what's possible with personal tokens, and we can definitely read org data with them: 
image

tap_github/repository_streams.py
Ry-DS: Sounds good. I updated the partioning keys like you mentioned here: [0b3f52f](https://github.com/MeltanoLabs/tap-github/pull/93/commits/0b3f52fd4a44d5f999bbb054731a6f398abfd81b). I left in `project_id` with the logic that there aren't many projects per project. Let me know if you see any issues with this :+1:

tap_github/tests/test_tap.py
laurentS: To address 1. yes, we can put multiple repos in the config, all from MeltanoLabs if needed. I can't think of any problems with this in CI.

laurentS:
> > > @Ry-DS It looks like the API is indeed returning them as strings, but they are clearly integers... Can we force the conversion? Do you get an error?
> > > -> https://api.github.com/repos/facebook/react/events
> > 
> > 
> > It makes sense that it's a string as the number is quite large (20 billion) which definitely won't fit in a 32bit signed integer (max 2 billion). I think it's the right move to keep it as strings (it would error out otherwise).
> 
> I've logged an issue in the SDK to keep this discussion going. For now, I agree it sounds like preserving as a string type is the "safe" option.
> 
> https://gitlab.com/meltano/sdk/-/issues/351

In our use-case, we pipe the output from this tap into [target-postgres](https://github.com/oviohub/target-postgres/) which actually [converts `integer` into `bigint`](https://github.com/oviohub/target-postgres/blob/master/target_postgres/postgres.py#L859) which is an 8 byte integer in pg, so ranging way beyond 2 billion.
I don't know if there is a standard conversion across taps/targets...

.github/workflows/test_tap.yml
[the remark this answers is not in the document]
ericboucher: The only question is - does the caching work between python versions? I am not sure how the matrix handles caching tbh. If it does then we can remove the fail-fast option. Otherwise we should keep it

tap_github/streams.py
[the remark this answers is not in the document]
ericboucher: I meant that potentially you could create a list called `repository_streams_list` in repository_strems.py and simply import it here. So this file would get simpler because the import would become:

import `repository_streams_list` from repository_streams

And the list below would just be `repository_streams_list`.

Same for the other groups. wdyt @Ry-DS ?

tap_github/tests/test_tap.py
ericboucher: So maybe the easiest path forward is for you (@aaronsteers ) or someone else on the MeltanoLab org to add a personal access token to this CI?

tap_github/streams.py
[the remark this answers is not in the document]
Ry-DS: I tried this out but tbh not a fan because the streams files are already very large, and adding another large block at the bottom doesn't feel like it helps. Think it's fine as it is now as there is now a central location to import all streams. Thoughts?

laurentS (approved):

laurentS:
There's something very weird with pytest, in the 3.10 environment, it says it's running python 3.8.10:
image
I suspect it came from the cached environment, despite the key being per python version, because I can't think of another reason for this to happen.

tap_github/tests/test_tap.py
```diff
@@ -22,6 +26,27 @@ def test_validate_repo_list_config(repo_list_config):
     assert partitions == repo_list_context
 
 
+def alternative_sync_chidren(self, child_context: dict) -> None:
```
laurentS, line 29: Maybe worth adding a short docstring here?

laurentS: Also, I'm wondering about how stable that upstream method is. If its logic changes, we'll have to amend the code here as well to reflect that. Seeing how short it is, it's quite unlikely though...
As an alternative option, we could fiddle with the singer catalog at the start of the test run to deselect this stream if the conditions are not right.

ericboucher: Added a docstring. I think it should be pretty stable and this gives us more control for future streams I think.

tap_github/tests/test_tap.py
edgarrmondragon: `ORG_LEVEL_TOKEN` is now available 🙂

PR #97, opened 2022-03-22
Correct error handling for http 401 codes

laurentS:
This PR corrects how the tap handles HTTP 401 error codes from the API:
the `Unauthorized` string appears in `response.reason` not `response.content`.

It also adds a fatal error if the token in invalid as there is no point retrying in that case.

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -197,10 +197,16 @@ class GitHubRestStream(RESTStream):
             # The GitHub API randomly returns 401 Unauthorized errors, so we try again.
             if (
                 response.status_code == 401
-                and "unauthorized" in str(response.content).lower()
+                and "unauthorized" in str(response.reason).lower()
             ):
+                # if the token is invalid however, we are also told about it
+                if "bad credentials" in str(response.content).lower():
+                    raise FatalAPIError("Auth token provided is invalid")
                 raise RetriableAPIError(msg)
 
+            # all other errors are fatal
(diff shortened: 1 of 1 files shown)


tap_github/client.py
```diff
@@ -197,10 +197,16 @@ def validate_response(self, response: requests.Response) -> None:
             # The GitHub API randomly returns 401 Unauthorized errors, so we try again.
             if (
                 response.status_code == 401
-                and "unauthorized" in str(response.content).lower()
+                and "unauthorized" in str(response.reason).lower()
```
ericboucher, line 200: how about changing that to: `and not "bad credentials" in str(response.content).lower()` and remove the changes below?

Also, it seems that "content" is more informative than "reason", should we change the messages to this?

```
f"{response.status_code} Client Error: "
f"{response.content} for path: {full_path}"
```

ericboucher: or both content and reason?

tap_github/client.py
```diff
@@ -182,7 +182,7 @@ def validate_response(self, response: requests.Response) -> None:
         if 400 <= response.status_code < 500:
             msg = (
                 f"{response.status_code} Client Error: "
-                f"{response.reason} for path: {full_path}"
+                f"{response.content} for path: {full_path}"
```
ericboucher, line 185: @laurentS maybe change it for `tolerated errors` above as well?

ericboucher:
:shipit:

PR #100, opened 2022-03-31
Avoid printing out "error" in tolerated status code message

ericboucher:
With the current logging implementation, we end up printing the word "error" for tolerated status codes.

Eg. `502 Tolerated Status Code: b'{\n  "message": "Server Error"\n}\n' (Reason: Bad Gateway) for path: /repos/kubevirt/kubevirt/issues/comments`

Instead, I propose to remove the content print and simplify the message to:

`502 Tolerated Status Code (Reason: Bad Gateway) for path: /repos/kubevirt/kubevirt/issues/comments`


laurentS (approved):
LGTM

PR #109, opened 2022-04-12
Add a config option to exclude streams by name

ericboucher:


tap_github/tests/test_tap.py
```diff
@@ -77,3 +78,26 @@ def test_get_a_repository_in_repo_list_mode(capsys, repo_list_config):
...
+
+    config2 = {**config1, "exclude": ["collaborators", "events", "readme"]}
+    tap2 = TapGitHub(config=config2)
+    assert tap1.catalog_dict["streams"] != tap2.catalog_dict["streams"]
+
+    tap2.sync_all()
+    captured = capsys.readouterr()
+    # Verify we got the right number of records (one per repo in the list)
+    assert captured.out.count('{"type": "RECORD", "stream": "repositories"') == 1
+    assert captured.out.count('{"type": "RECORD", "stream": "events"') == 0
```
edgarrmondragon, line 103: Should the other excluded streams be tested too?

tap_github/tap.py
```diff
@@ -56,8 +58,31 @@ class TapGitHub(Tap):
...
+            :class:`singer_sdk.helpers._singer.Catalog`.
+        """
+        excluded_streams = self.config.get("exclude", [])
+
+        def get_stream_entry(stream: Stream):
+            singer_catalog_entry = stream._singer_catalog_entry
+            singer_catalog_entry.metadata.root.selected = (
+                stream.tap_stream_id not in excluded_streams
+            )
+            return (stream.tap_stream_id, singer_catalog_entry)
```
edgarrmondragon, line 82: Should this return an instance of `CatalogEntry`?

edgarrmondragon (commented):
@ericboucher code looks good, but I'm not sure how this interacts with the catalog or which one takes precedence.

The code suggests that even if a stream is deselected in the catalog, by not being in the `excluded` array, it is actually selected.

tap_github/tap.py
ericboucher: We could, I just re-used the minimal object that the default function uses

ericboucher:
@edgarrmondragon great catch, I switched to using this, wdyt?

```
            singer_catalog_entry.metadata.root.selected = (
                singer_catalog_entry.metadata.root.selected
                and (stream.tap_stream_id not in excluded_streams)
            )
```

aaronsteers:
> @ericboucher code looks good, but I'm not sure how this interacts with the catalog or which one takes precedence.

@ericboucher - Can you say a bit more about the driver for this feature? Presumably if using Meltano there's a handy `--select` and `--exclude` CLI option, and other orchestrators like Stitch also have UIs for exclusion/inclusion. Why not rely on catalog selection logic here?

ericboucher:
We actually use the tap outside of Meltano or Stitch at the moment so it would be a convenient feature for us, but not critical

For example, the tap is used here to test the postgres target. And having this feature would have made it easier to exclude streams that require special access without inputing a full catalog schema. See https://github.com/datamill-co/target-postgres/pull/213

TLDR, it would be convenient for us and I think the current implementation does not interact negatively with the standard "catalog" way.

But I am open to suggestions, and we could also close this for now if you prefer. cc @laurentS

ericboucher:
Taking silence as a "let's close this for now"

Issue #110, opened 2022-04-21
Tap-github allows inputting buggy data into its output


laurentS:
It is possible to "corrupt" data coming out of the tap by inputting incorrectly cased values in the config file. While it does not directly affect the tap itself, it can lead to problems downstream (we've just hit this problem).
I am not sure what the correct solution is, so I'm opening this ticket for discussion and awareness.

In config, set `"repositories": ["meltanolabs/tap-github"]` (lowercase `M` and `L` on purpose) and run the tap deselecting the `repositories` stream, and keeping at least one child stream, `issues` for instance.

The tap will run fine, since the github api will handle the case inconsistencies in the request and return data containing `MeltanoLabs` instead of `meltanolabs`.

But because of these lines https://github.com/MeltanoLabs/tap-github/blob/0851e80578bbb17f69faa4c6b2220d30ed1802c5/tap_github/repository_streams.py#L103-107 the output from the tap will contain `org=meltanolabs` but `full_name=MeltanoLabs/tap-github` (note the difference in upper/lowercase chars). If you use the `org` and `repo` columns to join records from different streams together, you can end up with inconsistent values which are hard to reconcile.

The obvious solution to this problem is to avoid incorrectly cased input in the config, but this is fairly fragile. Another option would be to change the way the child streams get their `org/repo` values by parsing the `url` field for instance (which is received from github and matches the "official" repo name casing), but this seems fairly costly performance-wise.

Are there any best practices around this in other taps?

aaronsteers:
Hi, @laurentS. Thanks for raising. I definitely think this is worth fixing but I'm not sure as of yet what the best fix would be.

This is happening, I think, because `MeltanoLabs/tap-github` is _both_ a config value and _also_ a data element. While the source API may not be case sensitive, we certainly should expect the data to be.

PR #112, opened 2022-04-26
Fetch databaseId for repos and users when initializing streams

laurentS:
This PR aims to resolve a number of issues by adding the numeric repo id to the context of all streams in the tap.

We have been having issues with repos changing names, incorrect case in the config file, etc... leading to problems with joins downstream. We are using `org/repo` to connect the various streams together, but since those values are not stable (for instance if a repo is renamed), this can lead to issues.

To remedy this, we fetch the github `id` (or `databaseId` in their graphql API lingo) and add it to all streams via the child context. This `id` is stable across renames, which should help match records over time.

The PR adds a "preflight" request to the graphql API which fetches said `id` for all the repos in the config file (if `repositories` is indeed in the config). At the same time, it:
- gets the correct casing for mistyped names (so `meltanolabs/tap-github` is corrected to `MeltanoLabs/tap-github`, see #110)
- gets new names for repos that have been renamed (running a REST API call on `myorg/oldname` will return an HTTP 301, but we save this step here)
- filters out repos which don't exist and issues a warning in the logs for each of them. @aaronsteers you mentioned elsewhere that you'd be in favour of letting the tap error out, does the above seem like an acceptable trade-off to you? 

These changes should help increase the resilience of the tap to invalid manual input, as well as repo renames. If this logic seems ok to everyone, we can also apply it to the user streams, as username are likely to change more often than repo names.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,4 @@
 [pytest]
 markers =
     repo_list: mark a test as a repo_list.
+    username_list: make a test as using a list of usernames in config
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 markers =
-    repo_list: mark a test as a repo_list.
-    username_list: make a test as using a list of usernames in config
+    repo_list: mark a test as using a list of repos in config
+    username_list: mark a test as using a list of usernames in config
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -442,7 +442,8 @@ class EventsStream(GitHubRestStream):
         # the parent stream, e.g. for fork/parent PR events.
         row["target_repo"] = row.pop("repo", None)
         row["target_org"] = row.pop("org", None)
-        row["repo_id"] = context["repo_id"]
+        if context is not None:
+            row["repo_id"] = context["repo_id"]
         return row
 
     schema = th.PropertiesList(
@@ -813,7 +814,8 @@ class IssuesStream(GitHubRestStream):
         # replace +1/-1 emojis to avoid downstream column name errors.
         row["plus_one"] = row.pop("+1", None)
         row["minus_one"] = row.pop("-1", None)
-        row["repo_id"] = context["repo_id"]
+        if context is not None:
+            row["repo_id"] = context["repo_id"]
         return row
 
     schema = th.PropertiesList(
@@ -890,7 +892,8 @@ class IssueCommentsStream(GitHubRestStream):
 
     def post_process(self, row: dict, context: Optional[Dict] = None) -> dict:
         row["issue_number"] = int(row["issue_url"].split("/")[-1])
-        row["repo_id"] = context["repo_id"]
+        if context is not None:
+            row["repo_id"] = context["repo_id"]
         if row["body"] is not None:
             # some comment bodies include control characters such as \x00
             # that some targets (such as postgresql) choke on. This ensures
@@ -950,7 +953,8 @@ class IssueEventsStream(GitHubRestStream):
     def post_process(self, row: dict, context: Optional[Dict] = None) -> dict:
         row["issue_number"] = int(row["issue"].pop("number"))
         row["issue_url"] = row["issue"].pop("url")
-        row["repo_id"] = context["repo_id"]
+        if context is not None:
+            row["repo_id"] = context["repo_id"]
         return row
 
     schema = th.PropertiesList(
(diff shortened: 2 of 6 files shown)


edgarrmondragon (dismissed):

tap_github/repository_streams.py
```diff
@@ -186,10 +253,15 @@ class ReadmeStream(GitHubRestStream):
     state_partitioning_keys = ["repo", "org"]
     tolerated_http_errors = [404]
 
+    def post_process(self, row: dict, context: Optional[Dict] = None) -> dict:
+        row["repo_id"] = context["repo_id"]
+        return row
+
```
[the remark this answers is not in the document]
laurentS: That is a very good point! I'm not sure why I didn't do it at first :see_no_evil:

laurentS:
@edgarrmondragon is there a way to disable sonarcloud notifications? It's particularly annoying to get a github notification + email each time we're pushing a commit, to tell me it's green. I've looked around sonarcloud and their user forum, but I suspect this is configurable at admin level, from what I'm reading.

tap_github/repository_streams.py
```diff
@@ -57,17 +57,81 @@ def records_jsonpath(self) -> str:  # type: ignore
         else:
             return "$[*]"
 
+    def get_repo_ids(self, repo_list: List[Tuple[str]]) -> List[Dict[str, str]]:
```
ericboucher, line 60: While we are at it, I think we should move this function to a util and add a similar one for users (and maybe orgs)

tap_github/tests/test_tap.py
```diff
@@ -13,13 +13,25 @@
     "MeltanoLabs/tap-gitlab",
     "MeltanoLabs/target-athena",
 ]
+# the github repo ids that match the repo names above
+# in the same order
+repo_ids = [
```
ericboucher, line 18: let's call this `repo_list_2_ids`

tap_github/repository_streams.py
laurentS: done! I've added the same method on the `users` stream, and added a bunch of parametrized tests to try and cover all cases.
I've kept the methods in the classes as, although the structure of the code is the same, the 2 methods differ enough to not be worth trying to refactor things.

tap_github/repository_streams.py
```diff
@@ -57,17 +57,81 @@ def records_jsonpath(self) -> str:  # type: ignore
...
+
+            @property
+            def query(self) -> str:
+                chunks = list()
+                # there is probably some limit to how many items can be requested
+                # in a single query, but it's well above 1k.
+                for i, repo in enumerate(self.repo_list):
+                    chunks.append(
+                        f'repo{i}: repository(name: "{repo[1]}", owner: "{repo[0]}") '
+                        "{ nameWithOwner databaseId }"
```
ericboucher, line 88: Should we get the nodeId as welll?

tap_github/tests/__init__.py
```diff
@@ -11,9 +11,13 @@
     "api_calls_tests_cache",
     backend="sqlite",
     # make sure that API keys don't end up being cached
-    ignored_parameters=["Authorization"],
+    # Also ignore user-agent so that various versions of request
+    # can share the cache
+    ignored_parameters=["Authorization", "user-agent"],
```
ericboucher, line 16: Nice find! Any other parameters we should ignore?

laurentS: I hope not. I looked at the data passed to the cache_key function, and this seemed to be the only value that might be annoying.

tap_github/user_streams.py
```diff
@@ -26,16 +26,103 @@ def path(self) -> str:  # type: ignore
...
+                self.user_list = user_list
+
+            @property
+            def query(self) -> str:
+                chunks = list()
+                # there is probably some limit to how many items can be requested
+                # in a single query, but it's well above 1k.
+                for i, user in enumerate(self.user_list):
+                    chunks.append(
+                        f'user{i}: user(login: "{user}") ' "{ login databaseId }"
```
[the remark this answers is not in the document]
laurentS: I see what you mean. Yeah, we could try that. It won't make a massive difference, but it should be a bit less redundant. I'm always a bit cautious with factoring code out, not to make it so abstract that it becomes unreadable. I find the test functions I wrote quite messy already...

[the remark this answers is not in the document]
ericboucher: Let's keep it like that for now then, maybe open an issue to:
- factor the code out
- add organizations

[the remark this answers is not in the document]
ericboucher: The next question is: should we bump the version (even though it is backward compatible)?

edgarrmondragon:
> @edgarrmondragon is there a way to disable sonarcloud notifications? It's particularly annoying to get a github notification + email each time we're pushing a commit, to tell me it's green. I've looked around sonarcloud and their user forum, but I suspect this is configurable at admin level, from what I'm reading.

@laurentS I don't think it's possible at the moment to disable the PR comments without uninstalling SonarCloud completely: https://community.sonarsource.com/t/disable-github-merge-request-comments/18520

ericboucher (approved):

pytest.ini
```diff
@@ -1,3 +1,4 @@
 [pytest]
 markers =
     repo_list: mark a test as a repo_list.
+    username_list: make a test as using a list of usernames in config
```
ericboucher, line 4: make -> mark ?

Issue #113, opened 2022-04-28
Test tokens during initialization


ericboucher:
Add a test when loading tokens to make sure that they are valid and have not expired.

Expired tokens return an "Unauthorized" error which is a bit cryptic to interpret mid-stream.

Instead, I think we should validate tokens during the initialization phase. Other ideas?

PR #115, opened 2022-04-28
Test tokens during initialization

ericboucher:
Closing #113. Filtering out tokens that don't pass a simple test (get rate_limit).

### DIFF
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -82,10 +82,29 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
         # Get rate_limit_buffer
         rate_limit_buffer = self._config.get("rate_limit_buffer", None)
 
-        # Dedup tokens and create a dict of TokenRateLimit
+        # Dedup tokens and test them
+        filtered_tokens = []
+        for token in list(set(available_tokens)):
+            try:
+                now = datetime.now()
+                modified_since = now.strftime("%a, %d %b %Y %H:%M:%S GMT")
+                response = requests.get(
+                    url="https://api.github.com/repos/MeltanoLabs/tap-github/readme",
+                    headers={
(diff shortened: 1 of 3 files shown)


tap_github/authenticator.py
```diff
@@ -161,13 +161,34 @@ def prepare_tokens(self) -> Dict[str, TokenRateLimit]:
...
+                response = requests.get(
+                    url="https://api.github.com/repos/MeltanoLabs/tap-github/readme",
+                    headers={
+                        "Authorization": f"token {token}",
+                        "If-Modified-Since": modified_since,
+                    },
+                )
+                response.raise_for_status()
+                filtered_tokens.append(token)
+            except Exception as e:
```
laurentS, line 179: Could you catch the specific exception that `requests` raises in this case instead of a generic one?

tap_github/authenticator.py
```diff
@@ -161,13 +161,34 @@ def prepare_tokens(self) -> Dict[str, TokenRateLimit]:
...
 
-        # Dedup tokens and create a dict of TokenRateLimit
+        # Dedup tokens and test them
+        filtered_tokens = []
+        for token in list(set(available_tokens)):
+            try:
+                now = datetime.now()
+                modified_since = now.strftime("%a, %d %b %Y %H:%M:%S GMT")
+                response = requests.get(
+                    url="https://api.github.com/repos/MeltanoLabs/tap-github/readme",
```
laurentS, line 171: As an alternative, you could check the rate limit status of the token, which [does not count against quotas](https://docs.github.com/en/rest/rate-limit#get-rate-limit-status-for-the-authenticated-user).

laurentS (approved):
Just threw in a couple of thoughts, but otherwise, it looks good to me.

edgarrmondragon (approved):

Issue #120, opened 2022-05-17
Use graphql endpoint for stargazers


ericboucher:
The GitHub REST API makes it impossible to traverse the stargazers endpoint for large repos and limits how many pages we can query.

For example, querying `https://api.github.com/repositories/10270250/stargazers?page=500&per_page=100` will result in the following error:

```
{
    "message": "In order to keep the API fast for everyone, pagination is limited for this resource. Check the rel=last link relation in the Link response header to see how far back you can traverse.",
    "documentation_url": "https://docs.github.com/v3/#pagination"
}
```

And contrary to what the message suggests, querying the "last" link still gives an error...

To make things worse, the API does NOT allow any of the following params for this endpoint:
- "since"
- "sort"

A solution is to paginate through a Graphql version instead:

```
query { 
  repository(name: "react" owner: "facebook") { 
    stargazers(first: 100 orderBy: {field: STARRED_AT direction: DESC}) {
      edges {
        node {
          id
          login
        }
        starredAt
      }
    }
  }
}
```

PR #121, opened 2022-05-17
Paginate backwards for stargazers

ericboucher:
Temporary solution to avoid exploding API quotas because the stargazers stream does not provide "since" or "sort/direction" parameters

To mitigate this, we traverse backwards and try to exit early.

WARNING - this does not work for large repositories such as facebook/react, see #120

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -6,6 +6,9 @@ import requests
 from singer_sdk import typing as th  # JSON Schema typing helpers
 from singer_sdk.helpers.jsonpath import extract_jsonpath
 
+from dateutil.parser import parse
+from urllib.parse import parse_qs, urlparse
+
 from tap_github.client import GitHubGraphqlStream, GitHubRestStream
 from tap_github.schema_objects import (
     user_object,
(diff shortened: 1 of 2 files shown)


tap_github/repository_streams.py
```diff
@@ -1496,6 +1499,59 @@ def http_headers(self) -> dict:
...
+
+        # Use header links returned by the GitHub API.
+        parsed_url = urlparse(response.links[next_page_key]["url"])
+        captured_page_value_list = parse_qs(parsed_url.query).get("page")
+        next_page_string = (
+            captured_page_value_list[0] if captured_page_value_list else None
+        )
+        if next_page_string and next_page_string.isdigit():
+            print(next_page_key)
+            print(int(next_page_string))
```
laurentS, line 1550: Leftover debug?

laurentS (approved):
So this PR tries to emulate a `since` param on the endpoint? Are the results actually ordered by `starred_at`? What happens on repos with lots of stars? Does it just crash?

ericboucher:
Yes for large repositories it basically gets one page of data and then exists if there is a since parameter.
If there is no since parameter, it will do the same thing as before -> get all the data it can, which is not everything.

Eg for `faceook/react` we would get 40k out of 180k stargazers.

Note that this PR will probably be closed / useless if we move forward with #123

ericboucher:
Less useful thanks to #120

Issue #122, opened 2022-05-17
Use repo_id instead of [repo, org] as partition key and index when possible


ericboucher:
Thanks to the changes introduced by #112, we now have a more reliable column to use for children streams.

Note, this is potentially a breaking changes and might change the DB index structure.

ericboucher:
Similarly, use `user_id` instead os `username` whenever possible

ericboucher:
It could be interesting to link this to #95 and catch when we introduce breaking changes by changing partition keys, primary keys and such

aaronsteers:
@ericboucher - Can we bump the major version when this comes through? I imagine many (most?) usages will break with the new primary key.

PR #123, opened 2022-05-18
Add a Graphql implementation of the repo StargazersStream

ericboucher:
This PR is adding a Graphql implementation of the repo StargazersStream and solving #120
This also integrated backward pagination and early exit by faking a "since" parameter.

Open questions:
1. should we replace the existing stream completely? (I lean towards yes, but not sure if it is strictly backward compatible). If we go with yes, then we don't need #121 and can close it.
2. the graphql endpoint is missing gravatar_id for the user object, is it even worth mentioning? (I lean towards no)

Bonus - I made a few tweaks that seem to improve requests-caching


edgarrmondragon (approved):

ericboucher:
@edgarrmondragon any thoughts on questions 1 and 2 above?

edgarrmondragon:
> @edgarrmondragon any thoughts on questions 1 and 2 above?

@ericboucher the schema and replication keys look the same, so the rest and graphql versions only differ in the pagination order?

ericboucher:
The replication keys are different owner, repo -> repo_id @edgarrmondragon

tap_github/tests/fixtures.py
```diff
@@ -93,3 +95,28 @@ def organization_list_config(request):
         "organizations": organization_list,
         "rate_limit_buffer": 100,
     }
+
+
+def alternative_sync_chidren(self, child_context: dict) -> None:
+    """
+    Override for Stream._sync_children.
+    Enabling us to use an ORG_LEVEL_TOKEN for the collaborators sream.
```
laurentS, line 103:
```suggestion
    Enabling us to use an ORG_LEVEL_TOKEN for the collaborators stream.
```

laurentS (approved):

edgarrmondragon:
@ericboucher I see the primary key will change with the new graphql stream. What do you think of the following plan:

- Change the name of the new stream to something like `stargazers_graphql`
- Add a deprecation warning, in the readme and maybe in the code too for the REST version of the stream
- Deselect the new version by default (TBD how we handle that)
- Release new version with both streams

In a future release we remove the REST version and _consider_ renaming the graphql one.

cc @aaronsteers

ericboucher:
@edgarrmondragon not sure I'm a fan since the next release would be yet another breaking change by renaming this`stargazers_graphql` to `stargazers` for consistency.

To be honest, since the previous stream was working so poorly I think it's reasonable to introduce this one and override the previous one directly. Maybe we can add a warning in the code on this streams initialization? Is there a way to force the target to drop existing tables / update indexes?

ericboucher:
@edgarrmondragon @aaronsteers I followed your suggestion and used the same partition keys so that we can "safely" override the REST stream completely. Which I did in the end.

The breaking id changes should happen in https://github.com/MeltanoLabs/tap-github/issues/122

Please have a final look and let me know what you think.

cc @laurentS for final thoughts as well

tap_github/repository_streams.py
```diff
@@ -1517,6 +1520,96 @@ def post_process(self, row: dict, context: Optional[Dict] = None) -> dict:
     ).to_dict()
 
 
+class StargazersGraphqlStream(GitHubGraphqlStream):
+    """Defines 'UserContributedToStream' stream. Warning: this stream 'only' gets the first 100 projects (by stars)."""
+
+    name = "stargazers"
+    query_jsonpath = "$.data.repository.stargazers.edges.[*]"
+    primary_keys = ["user_id", "repo_id"]
```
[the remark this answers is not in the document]
aaronsteers: Scratch that. I see this breaks in #122, which makes this increment then non-breaking.

laurentS:
> To be honest, since the previous stream was working so poorly I think it's reasonable to introduce this one and override the previous one directly. Maybe we can add a warning in the code on this streams initialization? Is there a way to force the target to drop existing tables / update indexes?

I don't think there's any defined behaviour on the target side, each target can do pretty much anything with the data, at least from the singer spec perspective, which just defines the interface between tap and target.
But I'd agree with the view that the current stream is behaving poorly. Also, taking into account the number of stars on this repo (5, of which 2 are commenting on this PR and 1 is with Meltano), I'd think it's fair to assume the number of users is very low (and those actually using this stream even lower).
Considering the sort of problems an unstable PKey can bring (see #110 and #112), I'd be in favour of simply replacing the old stream with the new one (and stable keys), with a big fat warning in the logs on init. As an option, rename the old stream `stargazers_rest` and keep it with a deprecation warning for a bit longer, so that on the off chance that someone is actually using the stream, they have an easy fix.

ericboucher:
Is it easy to use a stream with another name? Eg. use stargazers_rest but still write to a `stargazers` table? Is this something `meltano-cli` provides @edgarrmondragon ?

edgarrmondragon:
> Is it easy to use a stream with another name? Eg. use stargazers_rest but still write to a `stargazers` table? Is this something `meltano-cli` provides @edgarrmondragon ?

@ericboucher the SDK comes with stream renaming out of the box: https://sdk.meltano.com/en/latest/stream_maps.html#aliasing-a-stream-using-alias

tap_github/repository_streams.py
```diff
@@ -1488,6 +1488,13 @@ class StargazersStream(GitHubRestStream):
     # GitHub is missing the "since" parameter on this endpoint.
     missing_since_parameter = True
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # TODO - remove warning with next release.
+        self.logger.warning(
+            "This stream is deprecated. Please use the Graphql version instead 'stargazers'."
+        )
```
edgarrmondragon, line 1496: @ericboucher this will be shown even if folks are not selecting (i.e. in the catalog) this stream, is that the okt?

ericboucher: hmm good catch where would you put it?

PR #124, opened 2022-05-18
Add a `throttle_requests_ms` parameter to slow down calls to the API

ericboucher:
Add a `throttle_requests_ms` parameter to slow down the API and avoid hitting GitHub's API secondary limits.

This parameter is optional and users can provide a number of milliseconds.


ericboucher:
Not sure it is as useful as we thought. @laurentS do you think it's worth merging? Or should we close this for now?

PR #127, opened 2022-05-19
Add scraping functions and dependentsStream for repositories

ericboucher:
Add scraping functions and add dependentsStream for repositories.


tap_github/repository_streams.py
```diff
@@ -1953,3 +1954,54 @@ def _write_schema_message(self):
...
+    state_partitioning_keys = ["repo_id"]
+
+    @property
+    def url_base(self) -> str:
+        return self.config.get("api_url_base", self.DEFAULT_API_BASE_URL).replace(
+            "api.", ""
+        )
+
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        """Get the scraped the response and return an iterator of result rows."""
```
laurentS, line 1976: typo?

pyproject.toml
```diff
@@ -16,6 +16,8 @@ singer-sdk = "^0.4.7"
 types-simplejson = "^3.17.2"
 types-python-dateutil = "^2.8.6"
 nested-lookup = "^0.2.23"
+bs4 = "^0.0.1"
+types-beautifulsoup4 = "^4.11.1"
```
laurentS, line 20: If these are optional as you suggest in the scraping file, maybe they could be specified as such? https://python-poetry.org/docs/pyproject/#extras

laurentS: actually, I think the types package should be dev-dependencies instead.

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
...
+    options = soup.find_all("a", class_="select-menu-item")
+    links = []
+    if len(options) > 0:
+        for link in options:
+            links.append(link["href"])
+    else:
+        links.append(response.url)
+
+    if verbose:
+        print(links)
```
laurentS, line 23: I'm not sure what the goal is here, but should this use the logger instead of print? Print will send its output to stdout which should be reserved for the actual records, so I think this would actually break the tap's function.

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for link in links:
+        yield from _scrape_dependents(f"https://github.com/{link}", verbose=verbose)
```
laurentS, line 26: Is there a way (and/or need) to get this base url from the stream method?

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
...
+                "Could not find star and fork info. Maybe the GitHub page format has changed?"
+            )
+
+        repos = [
+            {"name_with_owner": name, "stars": s, "forks": f}
+            for name, s, f in zip(repo_names, stars, forks)
+        ]
+
+        if verbose:
+            print(repos)
```
laurentS, line 63: Same comment as above

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
...
+        yield from _scrape_dependents(f"https://github.com/{link}", verbose=verbose)
+
+
+def _scrape_dependents(url, verbose=False):
+    # Optional dependency:
+    from bs4 import BeautifulSoup
+
+    while url:
+        if verbose:
+            print(url)
```
laurentS, line 35: same comment as above

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
...
+        response = requests.get(url)
```
laurentS, line 36: You should probably use a [requests.Session](https://2.python-requests.org/en/master/user/advanced/#session-objects) here to keep performance to a reasonable level. I don't think it makes sense to reuse the tap's session as the host is different though.

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
...
+        links.append(response.url)
+
+    if verbose:
+        print(links)
+
+    for link in links:
+        yield from _scrape_dependents(f"https://github.com/{link}", verbose=verbose)
+
+
+def _scrape_dependents(url, verbose=False):
```
laurentS, line 29:
```suggestion
def _scrape_dependents(url: str, verbose: bool = False) -> Iterable[dict[str, Any]]:
```
so that the code can be checked by mypy. This probably needs some imports: `from typing import Any, Iterable`

tap_github/scraping.py
```diff
@@ -0,0 +1,76 @@
+"""Utility functions for scraping https://github.com
+
+Inspired by https://github.com/dogsheep/github-to-sqlite/pull/70
+"""
+import requests
+import time
+
+
+def scrape_dependents(response: requests.Response, verbose=False):
```
laurentS, line 9:
```suggestion
def scrape_dependents(response: requests.Response, verbose: bool = False) -> Iterable[dict[str, Any]]:
```

laurentS (changes requested):
This is cool :)
Just one question: what are the rate limits that apply to scraping github web?

pyproject.toml
ericboucher: Making it optional seems complicated with an opt-in. I moved the type to "dev" but I think we can keep beautifulsoup4 like this.

tap_github/repository_streams.py
```diff
@@ -1953,3 +1954,54 @@ def _write_schema_message(self):
...
             super()._write_schema_message()
             self._schema_emitted = True
+
+
+class DependentsStream(GitHubRestStream):
+    """Defines 'dependents' stream."""
+
+    name = "dependents"
+    path = "/{org}/{repo}/network/dependents"
+    primary_keys = ["repo_id", "dependent_name_with_owner"]
```
ericboucher, line 1964: Do I need to also create `dependent_name_with_owner` at the top level or is the current implementation ok @edgarrmondragon ? (it's passed as `{ dependent: { name_with_owner } }`

laurentS (approved):

tap_github/repository_streams.py
ericboucher: @laurentS do you know?

laurentS: I don't know if this applies to all targets, but `target-postgres` translates nested fields as `dependent__name_with_owner` (note the double underscore), but I'm not sure if this is part of the signer spec, or just valid for this particular target.

ericboucher: Ok, to be safe I'll extract it as an independent column then

PR #131, opened 2022-05-19
Rotate auth token in backoff_handler

### LINKED ISSUE #129: Update rate limit handling to use new sdk feature
In the sdk v0.4.9, a new function was added through [this MR](https://gitlab.com/meltano/sdk/-/merge_requests/277) which improves the sdk-level handling of rate limits.

`tap-github`'s handling still leads to failures in some cases: with multiple tokens passed in config and enough remaining quota overall, the tap can still stop running with an error like:
```
name=backoff level=ERROR message=Giving up _request(...) after 5 tries (singer_sdk.exceptions.RetriableAPIError: 403 Client Error: b'{"message":"API rate limit exceeded for user ID xxxxxxx.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}' (Reason: Forbidden) for path: /repos/camicroscope/caMicroscope/languages)
```
The key part of this message (in our production logs) is that the user ID remains the same across the 5 retries, which is not expected.

It looks like the tap's authenticator correctly rotates the token, but the backoff decorated request continues retrying with the same token for a given request, because the authenticator passes the token to the request in [prepare_request](https://gitlab.com/meltano/sdk/-/blob/main/singer_sdk/streams/rest.py#L282) but the request headers are not updated with the new token until the next record.

While this work most of the time, if the buffer is less than the number of retries or some other process is depleting quotas in parallel, it is possible to perform the exact same query 5 times in a row, without actually changing the token.

The new `backoff_handler` should be able to fix this.

laurentS:
This PR fixes #129 to handle the edge case described in it.

The limit to this is that the token will be rotated on every `RetriableAPIError` with this code, which may not be desirable.
I don't see any obvious way to prevent this, short of throwing a different exception upon rate limit errors, which would require modifying `request_decorator`. Happy to hear any other ideas :)


tap_github/client.py
```diff
@@ -243,6 +243,13 @@ def post_process(self, row: dict, context: Optional[Dict[str, str]] = None) -> d
             row["repo_id"] = context["repo_id"]
         return row
 
+    def backoff_handler(self, details: dict) -> None:
+        """Handle retriable error by swapping auth token."""
+        self.logger.info("Retrying request with different token")
```
ericboucher, line 248: This is only for 403, no? Other retriable error do not need to this.

ericboucher (approved):

tap_github/client.py
laurentS: Yes, we only need this for rate limit errors, but as I mentioned in the PR, I didn't find a way to easily filter errors. Ideas welcome :)

ericboucher: Sorry I looked to quickly. And we don't have access to the status code on the backoff handler? I feel like this would be a good addition for this function in the SDK. Maybe we can merge like this and open an issue on the SDK and our tap?

ericboucher:
I am SHOCKED that the backoff library doesn't return the response or even just its status code...

Anyways, investigating this I found out that we were missing a key component, actually updating to the SDK that has this new featured 😉

ericboucher:
@laurentS looks like upgrading the SDK broke tests around `collaborators`, we probably need to change our implementation of `alternative_sync_chidren`

tap_github/client.py
laurentS: Yes, I think it would make sense to have some way of knowing why the backoff was needed. The reason would be determined in `validate_response`, and an easy fix could be to set a value on the stream object, which this handler could read and reset, but it feels a little hacky. Or is it possible to attach some value to the `RetriableAPIError` object? but then I'm not even sure the handler actually gets the exception object anyway.

ericboucher:
I opened - https://github.com/litl/backoff/issues/158

edgarrmondragon:
@ericboucher 

> Noting that the SDK should upgrade to `backoff 2.0.1` to get better typing @edgarrmondragon ;)

Yup, we're blocked by https://github.com/transferwise/pipelinewise-singer-python/pull/61

> I opened - https://github.com/litl/backoff/issues/158

What you want is already possible in backoff `2.0` with the (undocumented) `backoff.runtime`: https://github.com/litl/backoff/blob/92e7b6e910accee30118a6f3d2cef4c62caa38f5/tests/test_integration.py#L32-L37.

We ended up porting it to the SDK (because of the blocker above) as [`RESTStream.backoff_runtime`](https://sdk.meltano.com/en/latest/classes/singer_sdk.RESTStream.html#singer_sdk.RESTStream.backoff_runtime) and there's an example here: https://sdk.meltano.com/en/latest/code_samples.html#custom-backoff that accesses the response object via the `RetriableError` instance 🙂 

Wdyt?

laurentS:
I found a way to rotate the token only on relevant errors. It's probably not the most reliable solution, but should be fine as a temp fix until the PRs mentioned above are merged.

PR #132, opened 2022-05-19
Fix team parents

SBurwash:
# What 
Fixed issue with team parents

# Why
Changed from string -> Object

# Fix
Edited properties schema to respect new layout


.secrets/.gitignore
```diff
@@ -1,6 +0,0 @@
-# IMPORTANT! This folder is hidden from git - if you need to store config files or other secrets,
```
ericboucher, line 1: Is this change voluntary?

ericboucher (approved):

ericboucher:
Make sure your tests are passing. You simply need to reformat by running `poetry run black .` ;)

.secrets/.gitignore
SBurwash: Yes but can be removed; at our company we follow the philosophy of "only one gitignore"

PR #142, opened 2022-06-08
Add cost callback for rest calls

laurentS:
This PR implements a simple mechanism for keeping track of API "costs" using the mechanism in .


ericboucher:
@laurentS looks like the tests are not passing for Python 3.8

pyproject.toml
```diff
@@ -12,7 +12,7 @@ requests = "^2.25.1"
 # singer-sdk = "^0.4.9"
 # For local SDK dev:
 # singer-sdk = {path = "../singer-sdk", develop = true}
-singer-sdk = {git = "https://github.com/oviohub/meltano-sdk.git", rev = "8e505a074c1bca31bdc7bb3e2c8102fe95bcbe05"}
```
ericboucher: We don't have a pinned version that fits our need yet?

ericboucher: It would be nice to move back to a version of the upstream repo

laurentS: I'm planning to go back to a normal version once the sdk PR is merged, but this allows us to test this code early.

ericboucher (approved):

pyproject.toml
ericboucher: Should we wait to merge?

laurentS: Yes, I would wait until the sdk is officially supporting the cost logging logic, then we can update and merge this. But for now, this branch allows testing and seeing if the result is useful/usable.

ericboucher: For reference - https://github.com/meltano/sdk/pull/704

ericboucher: @laurentS the PR above has been merged :)

Issue #128, opened 2022-05-19
Explore fixes for error in issue_comments - ValueError: invalid literal for int() with base 16: b''


ericboucher:
Investigate a bit more why the issues/comment is raising the following error `ValueError: invalid literal for int() with base 16: b''`

eg.

<img width="1027" alt="Screen Shot 2022-05-19 at 1 05 17 AM" src="https://user-images.githubusercontent.com/16843267/169212126-23f2a0df-d2ed-4eb5-ac3a-9b5ec2450768.png">

Possible solutions to explore, patching urllib3: https://github.com/psf/requests/issues/4248

ericboucher:
I was able to reproduce the server error on issue/comments directly with:

https://api.github.com/repos/hashicorp/terraform-provider-aws/issues/comments?sort=updated&direction=asc&since=2020-06-22T11:30&per_page=100

The problem I think is that the since parameter is too expensive for large repos. I see two primary options:
1. Act as if the since parameter does not work and paginate in descending order with an early stop instead
2. Try to use a graphql request instead

@laurentS wdyt?

laurentS:
> The problem I think is that the since parameter is too expensive for large repos. I see two primary options:
> 
>     1. Act as if the since parameter does not work and paginate in descending order with an early stop instead
> 
>     2. Try to use a graphql request instead

Option 3. Switch to a different REST endpoint that lists comments for a single issue, as opposed to the entire repo. This would be more "quota hungry" on the `core` quota, so graphql is likely a better option there, if it works.

I'd try whichever is the least work between the 2 options you suggest, and see if it works. This is not properly documented on their side, so short of trying, we probably won't know.

PR #148, opened 2022-06-22
Catch server error in issue comments

ericboucher:
Temporary alleviation of #128


tap_github/repository_streams.py
```diff
@@ -891,7 +891,16 @@ def get_records(self, context: Optional[Dict] = None) -> Iterable[Dict[str, Any]
             self.logger.debug(f"No comments detected. Skipping '{self.name}' sync.")
             return []
 
-        return super().get_records(context)
+        # Getting issue comments is sometimes too expensive on large repos and results server errors.
+        # To avoid crashing the tap completely, we wrap the get_records in a try/except.
+        try:
+            return super().get_records(context)
+        except ValueError:
+            self.logger.warning(
```
laurentS, line 899: Is it worth filtering on the error message itself (if at all possible), so as to ignore only these specific errors, and not ones we actually want to crash on?

tap_github/repository_streams.py
```diff
@@ -891,7 +891,16 @@ def get_records(self, context: Optional[Dict] = None) -> Iterable[Dict[str, Any]
             self.logger.debug(f"No comments detected. Skipping '{self.name}' sync.")
             return []
 
-        return super().get_records(context)
+        # Getting issue comments is sometimes too expensive on large repos and results server errors.
```
laurentS, line 894:
```suggestion
        # Getting issue comments is sometimes too expensive on large repos and results in server errors.
```

laurentS: Maybe worth linking to the relevant github api docs?

laurentS (approved):

tap_github/repository_streams.py
ericboucher: I could not really find a github doc for that, did you?

tap_github/repository_streams.py
ericboucher: I added a string check, not sure if it's the best way though

laurentS:
@ericboucher do we still need this PR?

tayloramurphy:
@ericboucher what's the status on this PR?

ericboucher:
I think we can close for now

PR #150, opened 2022-06-22
Bypass since parameter in issue_comments to avoid server errors

### LINKED ISSUE #128: Explore fixes for error in issue_comments - ValueError: invalid literal for int() with base 16: b''

ericboucher:
Fixes #128

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -84,13 +84,13 @@ class GitHubRestStream(RESTStream):
         # Unfortunately endpoints such as /starred, /stargazers, /events and /pulls do not support
         # the "since" parameter out of the box. So we use a workaround here to exit early.
         # For such streams, we sort by descending dates (most recent first), and paginate
-        # "back in time" until we reach records before our "since" parameter.
+        # "back in time" until we reach records before our "fake_since" parameter.
         request_parameters = parse_qs(str(urlparse(response.request.url).query))
         # parse_qs interprets "+" as a space, revert this to keep an aware datetime
         try:
             since = (
-                request_parameters["since"][0].replace(" ", "+")
-                if "since" in request_parameters
+                request_parameters["fake_since"][0].replace(" ", "+")
+                if "fake_since" in request_parameters
                 else ""
             )
         except IndexError:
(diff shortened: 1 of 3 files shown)


tap_github/repository_streams.py
```diff
@@ -882,6 +882,10 @@ class IssueCommentsStream(GitHubRestStream):
     # we have gaps in our data
     tolerated_http_errors = [502]
 
+    # GitHub is not missing the "since" parameter on this endpoint.
+    # But it is too expensive on large repos and results in a lot of server errors.
+    missing_since_parameter = True
```
laurentS, line 887: I find this name a bit confusing. Should it be called something like `use_fake_since_parameter`? As your comment says, the `since` parameter is not actually missing.

tap_github/client.py
```diff
@@ -84,13 +84,13 @@ def get_next_page_token(
...
         # For such streams, we sort by descending dates (most recent first), and paginate
-        # "back in time" until we reach records before our "since" parameter.
+        # "back in time" until we reach records before our "fake_since" parameter.
         request_parameters = parse_qs(str(urlparse(response.request.url).query))
         # parse_qs interprets "+" as a space, revert this to keep an aware datetime
         try:
             since = (
-                request_parameters["since"][0].replace(" ", "+")
-                if "since" in request_parameters
+                request_parameters["fake_since"][0].replace(" ", "+")
```
laurentS, line 92: Is this not breaking other streams that actually support the `since` parameter?

laurentS (commented):
Just to be sure I'm following the logic here:
for this stream, you bypass the normal `since` parameter, and instead use our handmade "fetch backwards" mechanism because it should be cheaper server-side. Maybe worth being extra clear in the comments in both files, because this is not the most obvious logic :)

tap_github/client.py
ericboucher: the streams that support the since parameter don't need to exit early, because the since parameter takes care of the filtering server-side already :)

ericboucher:
@laurentS I tried to clean up the logic a bit, let me know what you think

laurentS (approved):

tap_github/repository_streams.py
laurentS: I see what you were doing. Makes sense. I was thinking of having 2 different vars, like `missing_since_parameter` and `use_fake_since_parameter` but then the code would be super ugly. I think this is fine. Nothing that a good comment cannot fix :)

tap_github/client.py
laurentS: Maybe worth a short comment in the code?

PR #152, opened 2022-06-27
Update replication_key state for descending streams

ericboucher:
A few limitations in the GitHub API force us to paginate a few streams in descending order. Since we cannot leverage the automated signpost updates from the SDK, this PR makes sure we properly update the state at the end of the run.

In addition, this PR switches the `commits` stream to use `use_fake_since_parameter` since this endpoint paginates in descending order.

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -54,6 +54,32 @@ class GitHubRestStream(RESTStream):
         headers["User-Agent"] = cast(str, self.config.get("user_agent", "tap-github"))
         return headers
 
+    def get_records(self, context: Optional[dict]) -> Iterable[Dict[str, Any]]:
+        """Return a generator of row-type dictionary objects.
+
+        Each row emitted should be a dictionary of property names to their values.
+
+        Args:
+            context: Stream partition or context dictionary.
+
+        Yields:
+            One item per (possibly processed) record in the API.
+        """
+        yield from super().get_records(context)
+
+        # Important - Update state for streams in descending order
+        if self.use_fake_since_parameter:
+            state = self.get_context_state(context)
+            if set(["replication_key_signpost", "replication_key"]).issubset(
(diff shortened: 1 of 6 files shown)


laurentS (approved):

tap_github/client.py
```diff
@@ -54,6 +54,32 @@ def http_headers(self) -> Dict[str, str]:
...
+        if self.use_fake_since_parameter:
+            state = self.get_context_state(context)
+            if set(["replication_key_signpost", "replication_key"]).issubset(
+                state.keys()
+            ):
+                record: Dict = {}
+                record[state["replication_key"]] = state["replication_key_signpost"]
+                self._increment_stream_state(
+                    latest_record=record,
+                    context=context,
```
edgarrmondragon, line 80: @ericboucher this will not run until `get_records` is actually iterated over, given the _lazy_ nature of generators.

ericboucher: You mean at the end of super.get_records()? If so that's the goal yes. Or am I missing something? And get_records() needs to be called in a special way? Should this live somewhere else then? @edgarrmondragon

ericboucher:
I am not entirely sure it is working as expected and don't really have a good way to test. Open to ideas @laurentS @edgarrmondragon :)

laurentS:
> I am not entirely sure it is working as expected and don't really have a good way to test. Open to ideas @laurentS @edgarrmondragon :)

@ericboucher I've added a test for your code. I *think* it does what you want. It's a bit clunky as I had to fake both the system time and the response from github, but might be a useful example for other corner cases we need to verify in the future. Feel free to adjust if I've missed anything!

laurentS:
This PR is probably not needed anymore with meltano/sdk#1164

ericboucher:
Solved by #164

PR #154, opened 2022-06-28
Leverage conditional requests to save API quotas

### LINKED ISSUE #23: Consider using conditional requests to check for changes
The GitHub API allows for conditional requests that do not count towards rate-limiting.

It could be interesting to leverage these and avoid making requests if they are not necessary.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests

ericboucher:
Fixes #23 

Add an "If-modified-since" header to try and leverage conditional requests from GitHub

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -2022,7 +2022,7 @@ class DependentsStream(GitHubRestStream):
 
         Mock a web browser user-agent.
         """
-        return {"User-agent": "Mozilla/5.0"}
+        return {"User-Agent": "Mozilla/5.0"}
 
     schema = th.PropertiesList(
         # Parent keys
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1988,7 +1988,10 @@ class WorkflowRunJobsStream(GitHubRestStream):
 
 
 class DependentsStream(GitHubRestStream):
-    """Defines 'dependents' stream."""
+    """Defines 'dependents' stream.
+
+    This stream scrapes the website instead of using the API.
+    """
 
     name = "dependents"
     path = "/{org}/{repo}/network/dependents"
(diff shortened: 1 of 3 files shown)


edgarrmondragon:
> is it reasonable to add a context-aware header in `get_url_params`? I couldn't really find a better way to do this but let me know if you have other ideas.

@ericboucher This is currently the best (and only!) way to pass values from context to headers. It does make sense to have better support in the SDK, though.

tap_github/repository_streams.py
```diff
@@ -2022,7 +2022,7 @@ def http_headers(self) -> dict:
 
         Mock a web browser user-agent.
         """
-        return {"User-agent": "Mozilla/5.0"}
+        return {"User-Agent": "Mozilla/5.0"}
```
laurentS, line 2025: Slightly off-topic for this PR, but GitHub [request](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required) that a clear app name be used, should this say `tap-github` instead?

laurentS (commented):
When you use the conditional requests header, the server might return an HTTP 304 status. Do we need to handle it in a particular way in the stream?

tap_github/repository_streams.py
ericboucher: This is mocking user http requests, not pinging the API so I think we actually don't want to give a clear app name 😄

laurentS: Hmm, I thought this was pretending to be a web browser, but the stream is actually pinging the API, no? :thinking:

ericboucher: No, this stream is a bit convoluted but it is actually pinging GitHub's website. The switch happens in `scrape_dependents` and we use beautifulsoup to parse the html

laurentS: Ok, I totally missed that. Just added a short comment to make it more obvious.

laurentS (approved):

PR #162, opened 2022-10-26
Add python 3.11 to CI workflows

laurentS:
Just adds a run with the newly released python 3.11. Hopefully everything should run smoothly :ski: !

### DIFF
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -20,7 +20,7 @@ jobs:
       ORG_LEVEL_TOKEN: ${{secrets.ORG_LEVEL_TOKEN}}
     strategy:
       matrix:
-        python-version: ["3.10", 3.9, 3.8, 3.7]
+        python-version: [3.11, "3.10", 3.9, 3.8, 3.7]
       # run the matrix jobs one after the other so they can benefit from caching
       max-parallel: 1
 


ericboucher (approved):

laurentS:
The tests are a bit cheeky (from the logs of the python 3.11 execution):
```
The currently activated Python version 3.11.0 is not supported by the project (<3.11,>=3.7.2).
Trying to find and use a compatible version. 
Using python3.8 (3.8.10)
```

laurentS:
This PR cannot be fixed until the sdk is upgraded to support python 3.11.
Right now `poetry update` fails with the following error:
```
The current project's Python requirement (>=3.7.2,<3.12) is not compatible with some of the required packages Python requirement:
  - singer-sdk requires Python >=3.7.1,<3.11, so it will not be satisfied for Python >=3.11,<3.12
  - singer-sdk requires Python >=3.7.1,<3.11, so it will not be satisfied for Python >=3.11,<3.12

Because no versions of singer-sdk match >0.12.0,<0.13.0 || >0.13.0
 and singer-sdk (0.12.0) requires Python >=3.7.1,<3.11, singer-sdk is forbidden.
So, because singer-sdk (0.13.0) requires Python >=3.7.1,<3.11
 and tap-github depends on singer-sdk (>=0.12.0), version solving failed.
```

Waiting on https://github.com/meltano/sdk/pull/1115

ericboucher:
@laurentS shouldn't the tests be failing then?

laurentS:
> @laurentS shouldn't the tests be failing then?

Sorry, I realise my messages were confusing. Basically CI passes because `poetry` falls back to python 3.8 in the 3.11 run. The error in my last message was from a run on my machine after adjusting the `pyproject.toml` file and trying to resolve dependencies again with `poetry update` (which does not happen in CI).

So I think we don't have too many options, wait for `sdk` which is waiting on `pyarrow`. Their build is passing as of 2h ago, so it shouldn't take too long before they push a new release.

ericboucher (changes requested):
Let's wait until we have a stable release for the SDK :)

laurentS:
> Let's wait until we have a stable release for the SDK :)

Done, using sdk `0.13.1` now, which includes python 3.11 support, despite what the release notes say.

ericboucher (approved):

PR #164, opened 2022-11-10
Fix broken state messaging in sdk

laurentS:
This PR updates the sdk to fix a bug in state messaging from the tap which prevented certain streams (such `issue_comments`) from ending with a valid state message, making it impossible to actually run them incrementally.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -639,45 +639,36 @@ python-versions = ">=2.5, !=3.0.*, !=3.1.*, !=3.2.*"
 
 [[package]]
 name = "singer-sdk"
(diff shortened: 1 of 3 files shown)


ericboucher:
I hope this fixes it and replaces https://github.com/MeltanoLabs/tap-github/pull/152 but I would be more convinced if we had a small test for it :)

TODO
- [ ] add a test for issue_comments and confirm that https://github.com/MeltanoLabs/tap-github/pull/152 is not needed anymore
- [ ] wait for an SDK release

ericboucher (approved):

PR #169, opened 2022-11-22
Update CI caching action

laurentS:
This might fix a bug in the action, from https://github.com/marketplace/actions/cache#v3:

> Fixed cache not working with github workspace directory or current directory.

Which might explain why the venv is cached, but not the api requests file?


laurentS:
This does not seem to have fixed the issue, seeing how slow the CI job is for this PR...

Issue #138, opened 2022-05-23
Create a function to get useful data about projects, not readily available through the API


ericboucher:
Let's create a stream to get useful data not readily available from the API.

- open issues (the api returns open_issues as open issues + open prs...)
- open prs
- number of dependents
- number of dependencies
- number of contributors
- ...

Any other ideas @laurentS?

We can piggy back on the the work done with #127

PR #170, opened 2022-11-29
Add extra metrics

ericboucher:
Fixes https://github.com/MeltanoLabs/tap-github/issues/138
Add extra metrics for repositories by scraping the main repository web page.

Note that the number of dependencies is a bit harder to get since we would need to visit another page or make an API query.
For example something like this and add up depenciesCount. Is it worth doing in here?

```
        return """
          query repositoryDependencies($repo: String! $org: String!) {
            repository(name: $repo owner: $org) {
              dependencyGraphManifests (first: 50 withDependencies: true) {
                totalCount
                nodes {
                  dependenciesCount
                }
              }
            }
            rateLimit {
              cost
            }
          }
	"""
```


tap_github/scraping.py
```diff
@@ -85,3 +87,67 @@ def _scrape_dependents(url: str, logger: logging.Logger) -> Iterable[Dict[str, A
...
+    try:
+        title = tag["title"]  # type: ignore
+        if isinstance(title, str):
+            title_string = cast(str, title)
+        else:
+            title_string = cast(str, title[0])
+        return int(title_string.strip())
+    except KeyError:
+        logger.debug(
+            f"Could not parse counter {tag}. Maybe the GitHub page format has changed?"
```
laurentS, line 106: The error handling here and below is not consistent, one just logs a debug message (likely to be silenced by a normal log level of `INFO`) the other raises an exception, although the likely underlying cause is similar (page format change). Is this on purpose?

tap_github/scraping.py
```diff
@@ -85,3 +87,67 @@ def _scrape_dependents(url: str, logger: logging.Logger) -> Iterable[Dict[str, A
...
+        )
+        return 0
+
+
+def scrape_metrics(
+    response: requests.Response, logger: Optional[logging.Logger] = None
+) -> Iterable[Dict[str, Any]]:
+    import re
+
+    from bs4 import BeautifulSoup
```
laurentS, line 116: any reason why these imports are in the function rather than at file level? `bs4` is already imported at the top anyway, and `re` is a standard module, so it shouldn't cause import problems.

tap_github/scraping.py
```diff
@@ -85,3 +87,67 @@ def _scrape_dependents(url: str, logger: logging.Logger) -> Iterable[Dict[str, A
...
+        prs = parse_counter(
+            soup.find("span", id="pull-requests-repo-tab-count"), logger
+        )
+    except IndexError:
+        raise IndexError(
+            "Could not find issues or prs info. Maybe the GitHub page format has changed?"
+        )
+
+    dependents = parse_counter(
+        getattr(soup.find(text=re.compile(" Used by ")), "next_element", None), logger
```
laurentS, line 133: Is it worth compiling these regexps at the top of the file so they're cached for all streams, instead of being parsed each time?

tap_github/scraping.py
ericboucher: Yes, the two other values are optional and may not appear on the page. I'll make it clearer through comments

tap_github/scraping.py
ericboucher: bs4 is not imported completely. This is how it was done before but I don't really mind, I agree that re can be moved to the top though

tap_github/scraping.py
ericboucher: I actually decided to raise an error in both cases since the case contributors or dependents "not found" is covered by tag = None

tap_github/scraping.py
laurentS: I tried to `git blame` it to figure out why we did it that way, but no obvious reason. Let's leave it like this, hopefully it won't have too much of an impact.

tap_github/scraping.py
laurentS: I checked in the docs, you can also [pass a simple string](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#a-string) to search instead of a regexp. I have no idea if either is faster.

laurentS (approved):

tap_github/scraping.py
laurentS: Using a string does not seem to work, so let's go with regexps!

PR #177, opened 2023-01-25
feat: use new SDK standard tap tests

kgpayne:


ericboucher (changes requested):
@kgpayne thanks or your PR!

As much as possible, we would like to keep this tap pointing to stable release versions of the SDK. Once your PR gets merged to the singer-sdk, we will gladly up the version here.

When suggesting a PR, please add some context and reasoning behind your changes and why they should be integrated.

kgpayne:
@ericboucher thanks for checking this out, and apologies that it wasn't left in a completely clear state. 'Draft' would have been appropriate as this effort bridged both before and after the release of the new test framework. We selected a few high-priority Taps/Targets based on Hub usage data for testing both backwards compatibility (hence the dependency on a pre-release version here) and to begin the work to migrate existing tests to the new framework.

As this repo already has pretty comprehensive tests, I think it fine to leave this closed for now to focus on other priority Taps/Targets with less test coverage 🙂👍

ericboucher:
Thanks for the additional context :)
And of course, feel free to re-open a PR to improve tests and test coverage in the future!

PR #180, opened 2023-02-13
fix: Prevent crash during cost calculation

laurentS:
This PR fixes a small bug in the cost calculation logic which caused the tap to crash in some cases for lack of the expected key in the server's response.

As the sdk calculates costs before iterating through records, where the main error handling logic lives, it would sometimes never reach that logic. This change prevents this and makes sure that the main error logic is reached, at the expense of "proper" cost calculation.

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -405,5 +405,12 @@ class GitHubGraphqlStream(GraphQLStream, GitHubRestStream):
     ) -> Dict[str, int]:
         """Return the cost of the last graphql API call."""
         costgen = extract_jsonpath("$.data.rateLimit.cost", input=response.json())
-        cost = next(costgen)
+        try:
+            cost = next(costgen)
+        except StopIteration:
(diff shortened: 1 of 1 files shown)


tap_github/client.py
```diff
@@ -405,5 +405,12 @@ def calculate_sync_cost(
     ) -> Dict[str, int]:
         """Return the cost of the last graphql API call."""
         costgen = extract_jsonpath("$.data.rateLimit.cost", input=response.json())
-        cost = next(costgen)
+        try:
+            cost = next(costgen)
```
ericboucher, line 409: How about using a default value like this?

https://stackoverflow.com/questions/14250184/default-value-for-next-element-in-python-iterator-if-iterator-is-empty

ericboucher (approved):

tap_github/client.py
laurentS: Nice! I didn't know `next` could do that.

ericboucher (approved):

PR #181, opened 2023-02-14
fix: Raise an api error on graphql errors

laurentS:
The tap (and the sdk) were not reporting any graphql errors (because annoyingly, graphql chooses to ignore good HTTP practice by returning a status 200 even on fatal errors).

This PR adds a minimal level of reporting by checking for `errors` in the response.

@edgarrmondragon The sdk does not seem to check for graphql errors at all, it simply defers to the REST streams `validate_response` which is not adapted to it. Is it worth adding something in the `sdk` in that respect?

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -99,15 +99,13 @@ class RepositoryStream(GitHubRestStream):
                 as we actually expect these in this stream when we send an invalid
                 repo name.
                 """
-                GitHubRestStream.validate_response(self, response)
-                rj = response.json()
-                if "errors" in rj:
-                    # simplify error handling by looking at the first error only
-                    err = rj["errors"][0]
-                    if err.get("type") != "NOT_FOUND":
-                        raise FatalAPIError(f"Graphql error: {err}", response)
+                try:
+                    super().validate_response(response)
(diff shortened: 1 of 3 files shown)


edgarrmondragon:
> @edgarrmondragon The sdk does not seem to check for graphql errors at all, it simply defers to the REST streams `validate_response` which is not adapted to it. Is it worth adding something in the `sdk` in that respect?

@laurentS Probably, yes. I wasn't aware of any standard or spec around errors in graphql, but if there is one, it makes sense to use it in the default `validate_response` implementation for GraphQL streams. I logged https://github.com/meltano/sdk/issues/1421.

tap_github/repository_streams.py
```diff
@@ -92,6 +92,20 @@ def query(self) -> str:
                     )
                 return "query {" + " ".join(chunks) + " rateLimit { cost } }"
 
+            def validate_response(self, response: requests.Response) -> None:
```
ericboucher, line 95: This looks ok. Though you could also imagine doing a try/except around your new graphql implementation. Wdyt?

ericboucher (approved):

aaronsteers:
@laurentS - Thanks very much for creating this!

Issue #183, opened 2023-03-02
The 'pull_number' field not being populated for the 'pull_request_commits' stream


scottmtp:
The 'pull_number' field not being populated for the 'pull_request_commits' stream.

When I run:
`meltano --log-level=debug run tap-github target-postgres`

In the log output I can see that the  pull_request_commits.pull_number field is selected, and the pull number appears in the context in the `Beginning full_table sync of 'pull_request_commits' with context:` line.

The `RECORD` log lines for pull_request_commits contain the other parent keys (org, repo, repo_id), but not pull_number.

scottmtp:
I'm able to work around this with this change:
```
diff --git a/tap_github/repository_streams.py b/tap_github/repository_streams.py
index e80c4f1..0701250 100644
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1351,6 +1351,10 @@ class PullRequestCommits(GitHubRestStream):
         ),
     ).to_dict()
 
+    def post_process(self, row: dict, context: Optional[Dict[str, str]] = None) -> dict:
+        if context is not None and "pull_number" in context:
+            row["pull_number"] = context["pull_number"]
+        return row
 
 class ReviewsStream(GitHubRestStream):
     name = "reviews"
```

However, I'm not sure if this is the correct fix.

ericboucher:
Great catch @scottmtp. Indeed, we would love a PR for this. Your change looks quite reasonable :)

ericboucher:
@edgarrmondragon any thoughts on why this part of the context is not forwarded to the record by default? @scottmtp's change looks reasonable but I'm wondering if we should look into a higher-level fix.

edgarrmondragon:
> @edgarrmondragon any thoughts on why this part of the context is not forwarded to the record by default? @scottmtp's change looks reasonable but I'm wondering if we should look into a higher-level fix.

@ericboucher Without diving deeper, my guess is that `state_partitioning_keys` is limiting the keys passed to `pull_requests_commits` from its parent stream.

PR #185, opened 2023-03-06
fix: adding pull_number field to 'pull_request_commits'

scottmtp:
Fix issue #183 by adding pull_number field to the 'pull_request_commits' stream.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1346,6 +1346,10 @@ class PullRequestCommits(GitHubRestStream):
         ),
     ).to_dict()
 
+    def post_process(self, row: dict, context: Optional[Dict[str, str]] = None) -> dict:
+        if context is not None and "pull_number" in context:
+            row["pull_number"] = context["pull_number"]
+        return row
 
 class ReviewsStream(GitHubRestStream):
     name = "reviews"
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1351,6 +1351,7 @@ class PullRequestCommits(GitHubRestStream):
             row["pull_number"] = context["pull_number"]
         return row
 
+
 class ReviewsStream(GitHubRestStream):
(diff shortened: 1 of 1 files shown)


tap_github/repository_streams.py
```diff
@@ -1348,6 +1348,11 @@ class PullRequestCommits(GitHubRestStream):
         ),
     ).to_dict()
 
+    def post_process(self, row: dict, context: Optional[Dict[str, str]] = None) -> dict:
```
ericboucher, line 1351: @scottmtp could you try adding `pull_number` to `state_partitioning_keys` instead and see if that fixes it?
Also, could you add a small test for this change?

Thanks for your PR! :)

ericboucher: If it doesn't work, make sure to call the parent `post_process` function to ensure that you are also passing the `repo_id`

scottmtp: Yes, adding `pull_number` to `state_partitioning_keys` does indeed fix the issue.

laurentS: This works, but beware of the implication: when you add `pull_number` to `state_partitioning_keys`, you're also expanding the state file to keep track of the state of each individual pull request in the stream, which might lead to a gigantic number of state bookmarks for bigger repos. Considering that PRs are "short lived" for most of them, I'm not sure it's worth keeping so much state overhead.

edgarrmondragon: @ericboucher yeah, Laurent is correct and it would also be a breaking change by making old state incompatible. `post_process` seems like the right approach 👍

ericboucher:
@scottmtp I ended up creating https://github.com/MeltanoLabs/tap-github/pull/187 to clean-up the way we use `post_process`. If this suits your need, feel free to approve my PR. Alternatively, you can pull my changes into your PR and we can merge yours :)

Issue #191, opened 2023-04-18
ValueError: not enough values to unpack (expected at least 1, got 0) in repository_streams


sicarul:
Hey folks! I'm having this issue with the tap, is there any way to mitigate it?

Thanks!

```
[2023-04-18, 12:05:44 UTC] {ecs.py:196} INFO - [2023-04-18, 12:05:38 UTC] 2023-04-18T12:05:38.397630Z [info     ] Traceback (most recent call last): cmd_type=elb consumer=False name=tap-github producer=True stdio=stderr string_id=tap-github
[2023-04-18, 12:05:44 UTC] {ecs.py:196} INFO - [2023-04-18, 12:05:38 UTC] 2023-04-18T12:05:38.397844Z [info     ]   File "/project/.meltano/extractors/tap-github/venv/bin/tap-github", line 8, in <module> cmd_type=elb consumer=False name=tap-github producer=True stdio=stderr string_id=tap-github
[2023-04-18, 12:05:44 UTC] {ecs.py:196} INFO - [2023-04-18, 12:05:38 UTC] 2023-04-18T12:05:38.398049Z [info     ]     sys.exit(cli())            cmd_type=elb consumer=False name=tap-github producer=True stdio=stderr string_id=tap-github
[2023-04-18, 12:05:44 UTC] {ecs.py:196} INFO - [2023-04-18, 12:05:38 UTC] 2023-04-18T12:05:38.398256Z [info     ]   File "/project/.meltano/extractors/tap-github/venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__ cmd_type=elb consumer=False name=tap-github producer=True stdio=stderr string_id=tap-github
[2023-04-18, 12:05:44 UTC] {ecs.py:196} INFO - [2023-04-18, 12:05:38 UTC] 2023-04-18T12:05:38.398499Z [info     ]     return self.main(*args, **kwargs) cmd_type=elb consumer=False name=tap-github producer=True stdio=stderr string_id=tap-github
[2023-04-18, 12:05:44 UTC] {ecs.py:196} INFO - [2023-04-18, 12:05:38 UTC] 2023-04-18T12:05:38.398708Z [info     ]   File "/project/.meltano/extractors/tap-github/venv/lib/python3.10/site-packages/click/core.py", line 1055, in main cmd_type=elb consumer=False name=tap-github producer=True stdio=stderr string_id=tap-github
[...]

PR #192, opened 2023-04-18
fix: Handle empty stargazers

sicarul:
When there are no stargazers in the next page function, it fails, this PR addresses it, fixing https://github.com/MeltanoLabs/tap-github/issues/191


edgarrmondragon (approved):

edgarrmondragon:
Thanks @sicarul! This LGTM and seems like a safe change so I'm gonna go ahead and merge, and then tag a patch release 🙂

PR #194, opened 2023-04-21
ci: Wait 2 minutes to start second tests attempt instead of 1 hour

edgarrmondragon:


laurentS:
@edgarrmondragon I guess the test failure above sort of confirms that we need more than 2 min wait time between CI attempts. The 1h wait time was added to ensure github rate limits are reset (their limits are N reqs/hour). We've been agonising over this, but haven't really figured out the solution.

The best option is probably making request caching work, but we haven't managed to do this. See https://github.com/MeltanoLabs/tap-github/pull/60 and #169

PR #195, opened 2023-04-23
Fix transformation for plus/minus one reactions

sicarul:
The current transformation happens at the wrong nesting level so it doesn't happen. Fixing to obtain +1 and -1 reactions.


laurentS (approved):
This seems to make sense, our data is full of `null`s where it probably shouldn't be. Thanks for the fix!

edgarrmondragon (approved):

edgarrmondragon:
@sicarul can you merge `main` into your branch?

Thanks!

Issue #119, opened 2022-05-16
request-cache is not working properly


ericboucher:
Tests keep hitting rate limiting issues, which suggests that our request-cache implementation is not working as expected during CI runs. It is working fine locally, which makes it hard to debug...

We should resolve this problem to help with our CI, making it faster and less frustrating

ericboucher:
Idea, store the cache file somewhere else? In S3, or at least somewhere that gives us a bit more control?

laurentS:
Actually, I think I misunderstood what is happening. The cache key is `api-cache-v4` which is python version agnostic.

But the cached file is 3 weeks old in the screenshot above, so our code ignores it (it [expires the cache after 24h](https://github.com/MeltanoLabs/tap-github/blob/main/tap_github/tests/__init__.py#L20)) so it hits the API again.

My guess at this point is that the cache works the first day we use it, but it then becomes stale and ignored by `requests-cache`. 2 options:
- change code to expire the cache after much longer than 24h. The problem being that tests use a `start_date` of more or less "today". So beyond the first day, it's likely that requests might change
- use a cache key based on the current date, which should make the cache file renew every 24h, in line with the cache expiry and api requests.

laurentS:
To expand on the above, this is all because cached files are read-only, so the cache file is not updated beyond its first creation time.
See https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache for details.

PR #207, opened 2023-05-12
fix: update ci caching logic

laurentS:
Fixes (hopefully) #119.

See details in the issue. Hopefully this PR's tests should already see the benefit of the cache.

### DIFF
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -72,5 +72,11 @@ jobs:
       id: retry_test_pytest
       if: steps.test_pytest.outcome=='failure'         # check the step outcome, wait and retry
       run: |
-        sleep 60m
+        # sleep as little as possible to reduce CI run time
+        # This assumes that REST quota is the one that caused problem
+        # (which is most likely/often the case)
+        target_ts=$(curl -s -H "Accept: application/vnd.github+json"  -H "Authorization: Bearer $GITHUB_TOKEN" -H "X-GitHub-Api-Version: 2022-11-28"  https://api.github.com/rate_limit | grep reset | head -n 1 | awk -F: '{ print $2 }')
+        current_ts=$(date +%s)
+        seconds_to_sleep=$(echo "$target_ts - $current_ts" | bc)
+        sleep $seconds_to_sleep
         LOGLEVEL=WARNING poetry run pytest --capture=no
--- a/.github/workflows/test_tap.yml
+++ b/.github/workflows/test_tap.yml
@@ -26,14 +26,22 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    - name: Get Date
+      id: get-date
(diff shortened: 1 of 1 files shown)


laurentS:
@edgarrmondragon I'm wondering if the github api token used for the CI here is also used somewhere else. The changes in this PR seem effective, the last run took 35 minutes, 28 of which were waiting for new quotas. But everything was consumed within about 10 minutes. By my calculation (running tests locally), the full CI run with 5 python versions should use about 2500 REST requests, and this didn't happen here, we had much less actually happening (only 1 python version on 1 PR, caching on the other). Are you able to check if the token's rate allowance is eaten up by something else? CI runs took 3-4hrs over the past few days, which makes it very hard to work on the repo.

edgarrmondragon:
@laurentS It's probably easiest to regenerate the token and document the permissions. Do you know the minimum set of permissions required for the tap to run CI?

ericboucher (approved):

edgarrmondragon:
> > @laurentS It's probably easiest to regenerate the token and document the permissions. Do you know the minimum set of permissions required for the tap to run CI?
> 
> @edgarrmondragon Yes, let's try this. From what I gather in the code, we use 2 env vars for testing:
> 
>     * `GITHUB_TOKEN` has `public_repo` permissions
> 
>     * `ORG_LEVEL_TOKEN` has `public_repo` and `read:org` permissions (this is used on some tests which are skipped if this value is not set in the env)
> 
> 
> That seems to be enough to get all tests to pass locally, so I suppose it should be the same with CI. It'd be great if you're able to rotate tokens to understand what is eating quotas so quickly.

[`GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) is generated for every run so it shouldn't be competing for resources.

I've updated `ORG_LEVEL_TOKEN` with read-only permissions for the MeltanoLabs org repos. Let me know if there's any issues, otherwise I'll open a PR to document the process for rotating the token.

laurentS:
> [`GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) is generated for every run so it shouldn't be competing for resources.

I was going to say the rate limits are per account, not per token, but it's actually worse than that (from [the docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting)):

> When using GITHUB_TOKEN, the rate limit is 1,000 requests per hour per repository.

I think this explains why we're hitting limits so often. My (very rough) calculation above was that we use ~500 REST calls per python version (before this PR). If caching worked perfectly, then we'd hit the limit just once per day, but considering different branches and how they don't share cache, it's likely that we hit problems more frequently. I'll see if I can modify the tests a bit to use fewer requests per run.

PR #218, opened 2023-06-23
fix: skip repos for which contributors list is too large

laurentS:
A quick and dirty PR to get around the `torvalds/linux` problem (and other similar repos) where the contributors list is so large that github gives up with a confusing error code.

To be updated once https://github.com/meltano/sdk/pull/1754 is merged/released.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1526,7 +1526,8 @@ class ContributorsStream(GitHubRestStream):
                 == "The history or contributor list is too large to list contributors for this repository via the API."
             ):
                 self.logger.info(
-                    "Skipping repo '%s' as contributors list is too large", self.name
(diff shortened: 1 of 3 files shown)


ericboucher (approved):

tap_github/repository_streams.py
```diff
@@ -1510,6 +1510,27 @@ class ContributorsStream(GitHubRestStream):
...
+    def validate_response(self, response: requests.Response) -> None:
+        """Allow some specific errors."""
+        if response.status_code == 403:
+            contents = response.json()
+            if (
+                contents["message"]
+                == "The history or contributor list is too large to list contributors for this repository via the API."
+            ):
+                self.logger.info(
+                    "Skipping repo '%s' as contributors list is too large", self.name
```
ericboucher, line 1529: "Skipping repo '%s'. The list of contributors is too large.", self.name

PR #230, opened 2023-09-15
feat: log when a repo is renamed

ericboucher:
Handle the case when a repo is renamed by logging the info.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -120,7 +120,9 @@ class RepositoryStream(GitHubRestStream):
                 if item == "rateLimit":
                     continue
                 try:
-                    org, repo = record[item]["nameWithOwner"].split("/")
+                    repo_full_name = "/".join(repo_list[int(item[4:])])
+                    name_with_owner = record[item]["nameWithOwner"]
+                    org, repo = name_with_owner.split("/")
                 except TypeError:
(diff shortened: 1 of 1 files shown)


tap_github/repository_streams.py
```diff
@@ -130,6 +132,14 @@ def validate_response(self, response: requests.Response) -> None:
                         "Removing it from list"
                     )
                     continue
+                # check if repo has moved or been renamed
+                if repo_full_name.lower() != name_with_owner.lower():
+                    # the repo name has changed, log some details, and move on.
+                    self.logger.info(
+                        f"Repository name changed: {repo_full_name} \t"
+                        f"New name: {name_with_owner}"
+                    )
```
edgarrmondragon, line 141: Nit: Use printf-style for logs

```suggestion
                    self.logger.info(
                        "Repository name changed: %s \tNew name: %s",
                        repo_full_name,
                        name_with_owner,
                    )
```

https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/

edgarrmondragon (approved):

tap_github/repository_streams.py
ericboucher: I feel like f-strings are a bit more readable in this case no? If Meltano has a strong feeling for `%s` I'll open an issue to change other instances of logs as well

PR #253, opened 2024-05-07
feat: Sets MAX_RESULTS_LIMIT to 1000 when RepositoryStream.path is /search/repositories only

merlindavies:
This PR introduces a minor updates to the RepositoryStream class such the `MAX_RESULTS_LIMIT` is applied only when the path is :
`/search/repositories/`,
and not when it's 
`/repos/{org}/{repo}`
or 
`/orgs/{org}/repos`.

These last two endpoints can handle results greater than 1000.


edgarrmondragon (approved):
Thanks @merlindavies!

PR #254, opened 2024-05-14
fix: releases replication key

TylerHillery:
I have updated the `replication_key` for GitHub releases to `created_at` field vs the `published_at` field. 

The `pubished_at` field can be `null` according to the [REST API endpoints for releases](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28)

This was causing the below error
```
{"TypeError: '>' not supported between instances of 'NoneType' and 'str'"}
```

`created_at` field can't be null so this should fix that.


edgarrmondragon (approved):
Makes sense to me. Thanks @TylerHillery!

PR #255, opened 2024-06-05
fix: Fix issue with the repo_id type for user streams being incorrect

BraedonLeonard:
The schema would report repo_id as a string, which caused validation errors when running since repo_id is actually an integer.


edgarrmondragon (approved):

edgarrmondragon:
This does make sense but it might introduce a breaking change :(

BraedonLeonard:
> This does make sense but it might introduce a breaking change :(

Do you have any ideas on how this change could get introduced without being a breaking change?

edgarrmondragon:
> > This does make sense but it might introduce a breaking change :(
> 
> Do you have any ideas on how this change could get introduced without being a breaking change? Or at least a similar change that solves the issue of schema validation failing for every message in the starred repo stream

On a second look, this should actually be more or less a safe change, and a necessary one. If a column in the destination exists with a TEXT type, then it should be happy to accept an integer literal and cast it automatically.

I'm merging it :)

PR #260, opened 2024-07-04
chore: Lint with Ruff

edgarrmondragon:


edgarrmondragon:
Let's do this piecemeal :)

PR #268, opened 2024-07-05
feat: create CustomPropertiesStream class

MindaugasN:
create CustomPropertiesStream class that fetches data from https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28


edgarrmondragon:
Seeing a 404 in the tests: `Custom properties only supported for organizations`.

Perhaps we should think of a way of skipping/deselecting this stream for non-org namespaces?

MindaugasN:
> Seeing a 404 in the tests: `Custom properties only supported for organizations`.
> 
> Perhaps we should think of a way of skipping/deselecting this stream for non-org namespaces?

Is there already an example with that? And how to tell if a namespace is org / non-org?

edgarrmondragon:
> > Seeing a 404 in the tests: `Custom properties only supported for organizations`.
> > Perhaps we should think of a way of skipping/deselecting this stream for non-org namespaces?
> 
> Is there already an example with that? And how to tell if a namespace is org / non-org?

@MindaugasN I think moving it here

https://github.com/MeltanoLabs/tap-github/blob/f41b66d5b51a3b9dc86ba0c9dcfd15d98b91df21/tap_github/streams.py#L118-L121

would be enough

MindaugasN:
I moved, but first of all, I am not sure, if I did that correctly.
What I see though, like this - it does not work for me anymore, `CustomPropertiesStream` should be under repository stream.
Another thing, I don't understand why pre-commit.ci fails. Do you have an idea?

edgarrmondragon:
> I moved, but first of all, I am not sure, if I did that correctly. What I see though, puting stream like this (under organizations) - it does not work for me anymore, `CustomPropertiesStream` should be under repository stream.

Can you say more about this? What do you mean it's not working anymore?

> Maybe you didn't mean that I should move `CustomPropertiesStream` under organization_streams.py?

Not really. The stream class can stay in `tap_github/repository_streams.py`.

> Another thing, I don't understand why pre-commit.ci fails. Do you have an idea?

I think you're missing a newline at the end of `tap_github/organization_streams.py`.

MindaugasN:
> Not really. The stream class can stay in tap_github/repository_streams.py.
Got it, that solved all my problems except one - test still fails :) 

How to read the message / debug it? I tried add and removing newline at the end of the file - it is still the same.

edgarrmondragon:
I think we're back at https://github.com/MeltanoLabs/tap-github/pull/268#issuecomment-2210215048 😅

PR #278, opened 2024-07-25
fix: Expand CollaboratorsStream primary keys

TrishGillett:
The collaborators stream pulls one row per selected repo per user who is a known collaborator of that repo, so `id` (the user ID) alone isn't sufficient to describe the grain.  This PR adds `repo` and `org` to the primary keys to capture the fact that there can be a row for each user x repo pair.

My motivation is that I'm getting an error `Query error: UPDATE/MERGE must match at most one source row for each target row ...` when trying to extract this stream sequentially with `target-bigquery` and the [upsert](https://hub.meltano.com/loaders/target-bigquery/#upsert-setting) setting turned on, and that sounds related to the fact that the column set as the primary key isn't distinctive.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -747,7 +747,7 @@ class LanguagesStream(GitHubRestStream):
 class CollaboratorsStream(GitHubRestStream):
     name = "collaborators"
     path = "/repos/{org}/{repo}/collaborators"
-    primary_keys = ["id"]
+    primary_keys = ["id", "repo", "org"]
     parent_stream_type = RepositoryStream
     ignore_parent_replication_key = True
     state_partitioning_keys = ["repo", "org"]


TrishGillett:
I think the stream has been working properly in a couple use cases:
- cases where people only tried to pull data about one repo
- cases where people didn't do any operations that were relying on the uniqueness of the primary key. For example, even when extracting for multiple repos my first run went fine, I only encountered an issue when trying to do a second run of the extractor and upsert it into the table generated by the first run.

I expect users' jobs to continue to work fine in either case without intervention. If you're only pulling one repo, the key was distinctive before and it'll still be distinctive now, and if you aren't doing anything that relies on the primary key then you should be unaffected.

edgarrmondragon:
> I think the stream has been working properly in a couple use cases:
> 
>     * cases where people only tried to pull data about one repo
> 
>     * cases where people didn't do any operations that were relying on the uniqueness of the primary key. For example, even when extracting for multiple repos my first run went fine, I only encountered an issue when trying to do a second run of the extractor and upsert it into the table generated by the first run.
> 
> 
> I expect users' jobs to continue to work fine in either case without intervention. If you're only pulling one repo, the key was distinctive before and it'll still be distinctive now, and if you aren't doing anything that relies on the primary key then you should be unaffected.

That makes sense. Thanks @TrishGillett!

PR #281, opened 2024-07-30
refactor: Authenticator refactoring (preparation for app token refreshing)

TrishGillett:
**Note: In this PR I am not changing functionality but just refactoring to make functionality changes possible in a future PR.** Essentially: [make the change easy, then make the easy change.](https://x.com/KentBeck/status/250733358307500032?lang=en)

My goal is to add github app token features, for example refreshing app tokens before they expire (they're only good for an hour and currently don't refresh), and allowing multiple app tokens

Currently, personal and app tokens are stored in the same list and treated interchangeably, so some structural changes are needed in order to handle them differently.  My plan is to develop PersonalTokenManager and AppTokenManager classes so the code for working with each token type can be built into its manager class.  In this PR I start by proposing to convert the TokenRateLimit class to a more general TokenManager, and I move functionality common to both token types there. I've added unit tests for the methods in TokenManager.

Follow up PRs will develop PersonalTokenManager and AppTokenManager and implement the new app token features.

### DIFF
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -2,9 +2,9 @@
 
 import logging
 import time
+from copy import deepcopy
 from datetime import datetime
 from os import environ
-from copy import deepcopy
 from random import choice, shuffle
 from typing import Any, Dict, List, Optional, Set
 
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -4,6 +4,7 @@ import logging
 import time
 from datetime import datetime
 from os import environ
+from copy import deepcopy
 from random import choice, shuffle
 from typing import Any, Dict, List, Optional, Set
 
@@ -219,8 +220,8 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
         )
 
     def get_next_auth_token(self) -> None:
-        token_managers = self.token_managers
         current_token = self.active_token.token if self.active_token else ""
+        token_managers = deepcopy(self.token_managers)
         shuffle(token_managers)
         for token_manager in token_managers:
             if (
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -215,7 +215,7 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
         self._config: Dict[str, Any] = dict(stream.config)
         self.token_managers = self.prepare_tokens()
         self.active_token: Optional[TokenManager] = (
-            choice(self.token_managers) if len(self.token_managers) else None
+            choice(self.token_managers) if self.token_managers else None
         )
 
     def get_next_auth_token(self) -> None:
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -22,7 +22,12 @@ class TokenManager:
     # - not consume all available calls when we rare using an org or user token.
     DEFAULT_RATE_LIMIT_BUFFER = 1000
 
-    def __init__(self, token: str, rate_limit_buffer: Optional[int] = None, logger: Optional[Any] = None):
+    def __init__(
+        self,
+        token: str,
+        rate_limit_buffer: Optional[int] = None,
+        logger: Optional[Any] = None,
+    ):
         """Init TokenManager info."""
         self.token = token
         self.logger = logger
@@ -143,7 +148,9 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
         if "auth_token" in self._config:
             personal_tokens.add(self._config["auth_token"])
         if "additional_auth_tokens" in self._config:
-            personal_tokens = personal_tokens.union(self._config["additional_auth_tokens"])
+            personal_tokens = personal_tokens.union(
+                self._config["additional_auth_tokens"]
+            )
         else:
             # Accept multiple tokens using environment variables GITHUB_TOKEN*
             env_tokens = {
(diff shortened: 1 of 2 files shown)


tap_github/authenticator.py
```diff
@@ -113,25 +140,37 @@ def generate_app_access_token(
 class GitHubTokenAuthenticator(APIAuthenticatorBase):
     """Base class for offloading API auth."""
 
-    def prepare_tokens(self) -> Dict[str, TokenRateLimit]:
+    def prepare_tokens(self) -> List[TokenManager]:
         # Save GitHub tokens
-        available_tokens: List[str] = []
+        rate_limit_buffer = self._config.get("rate_limit_buffer", None)
+
+        personal_tokens: Set[str] = set()
```
TrishGillett, line 147: Made this a set so we don't need to worry about deduplicating later.

tap_github/authenticator.py
```diff
@@ -41,7 +47,28 @@ def update_rate_limit(self, response_headers: Any) -> None:
         self.rate_limit_reset = int(response_headers["X-RateLimit-Reset"])
         self.rate_limit_used = int(response_headers["X-RateLimit-Used"])
 
-    def is_valid(self) -> bool:
+    def is_valid_token(self) -> bool:
```
TrishGillett, line 50: Some name swapping here:
- Renaming `is_valid` to `has_calls_remaining` because it's actually just checking the call budget.
- Naming this new function `is_valid_token` because it is actually testing tokens' validity.

tap_github/authenticator.py
```diff
@@ -113,25 +140,37 @@ def generate_app_access_token(
...
-            ]
+            }
             if len(env_tokens) > 0:
                 self.logger.info(
                     f"Found {len(env_tokens)} 'GITHUB_TOKEN' environment variables for authentication."
                 )
-                available_tokens = env_tokens
+                personal_tokens = env_tokens
+
+        token_managers: List[TokenManager] = []
```
TrishGillett, line 167: This list replaces the token map [here](https://github.com/MeltanoLabs/tap-github/pull/281/files#diff-0318b29f82fbeb258f98e6f2c373e09ebc2869589302977b4398ab4f50347e9bL186) that was a dict mapping token strings to their TokenRateLimit objects. This was unnecessary since the TokenRateLimit already knew the value of its token, and would be problematic when implementing token refreshing since the token value will change hourly in the app token case.

tap_github/authenticator.py
```diff
@@ -196,18 +213,21 @@ def __init__(self, stream: RESTStream) -> None:
...
 
     def get_next_auth_token(self) -> None:
-        tokens_list = list(self.tokens_map.items())
+        token_managers = self.token_managers
         current_token = self.active_token.token if self.active_token else ""
-        shuffle(tokens_list)
-        for _, token_rate_limit in tokens_list:
-            if token_rate_limit.is_valid() and current_token != token_rate_limit.token:
-                self.active_token = token_rate_limit
+        shuffle(token_managers)
```
TrishGillett, line 224: This line is being flagged as suspicious by sonarcloud.  All it's doing is shuffling the list of managers so we have a chance to pick any token manager next, perfect randomization is not critical.

edgarrmondragon: I've marked it as safe in the sonarcloud UI. Mildly annoying, sorry 😅.

tap_github/authenticator.py
edgarrmondragon: Thanks!

tap_github/authenticator.py
```diff
@@ -196,18 +213,21 @@ def __init__(self, stream: RESTStream) -> None:
         self.logger: logging.Logger = stream.logger
         self.tap_name: str = stream.tap_name
         self._config: Dict[str, Any] = dict(stream.config)
-        self.tokens_map = self.prepare_tokens()
-        self.active_token: Optional[TokenRateLimit] = (
-            choice(list(self.tokens_map.values())) if len(self.tokens_map) else None
+        self.token_managers = self.prepare_tokens()
+        self.active_token: Optional[TokenManager] = (
+            choice(self.token_managers) if len(self.token_managers) else None
```
edgarrmondragon, line 218: fwiw I think we can simplify this

```suggestion
            choice(self.token_managers) if self.token_managers else None
```

tap_github/authenticator.py
edgarrmondragon, line 224: This reorders the list in place, right? Does that affect the `self.token_managers`? Is that the intended behavior?

Totally OK if it's _yes_ to all, just wanna double check 🙏

tap_github/tests/test_authenticator.py
edgarrmondragon, line 1: This is really great!

tap_github/authenticator.py
TrishGillett: Good catch!  The order of `self.token_managers` shouldn't matter in theory since the order isn't currently used anywhere (until now it was a dict after all), but changing it also isn't intentional and could trip up a future dev.  I'll confirm whether `self.token_managers` is getting changed and tweak if so. 🙏

TrishGillett: Yup, it was shuffling `self.token_manager`.  Pushed a commit adding a `deepcopy`.

edgarrmondragon: Awesome, thanks for confirming and updating!

edgarrmondragon (approved):
Thanks @TrishGillett!

PR #284, opened 2024-08-03
fix: Tokens in config are no longer ignored when there are tokens in the environment, part of refactoring in preparation for app token refreshing

TrishGillett:
This PR continues the work started in https://github.com/MeltanoLabs/tap-github/pull/281, completing the major refactor. This is accomplished by creating PersonalTokenManager and AppTokenManager classes, moving logic to them, and adding tests.

After this PR, I'll be able to easily implement refreshing for app tokens and other app token related features.

In additional to the added unit tests, I have tested running an extractor on top of the proposed code, in a couple scenarios:

with auth_token? | with additional_auth_tokens? | with personal tokens in env? | with app token? | # tokens that should be used | # tokens used in test run
--|--|--|--|--|--
no | no | no | yes | 1 | 1 ✅ 
yes | yes (1) | no | yes | 3 | 3 ✅ 
no | yes (2) | no | yes | 3 | 3 ✅ 

I am not easily able to test the case of personal tokens being detected from the environment, just because changing the name of environment variables is not simple in my set up.

### DIFF
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -102,7 +102,7 @@ def generate_jwt_token(
     github_app_id: str,
     github_private_key: str,
     expiration_time: int = 600,
-    algorithm: str = "RS256"
+    algorithm: str = "RS256",
 ) -> str:
     actual_time = int(time.time())
 
@@ -189,12 +189,16 @@ class AppTokenManager(TokenManager):
                 '":app_id:;;-----BEGIN RSA PRIVATE KEY-----\\n_YOUR_P_KEY_\\n-----END RSA PRIVATE KEY-----"'
             )
 
-        self.token, self.token_expires_at = generate_app_access_token(self.github_app_id, self.github_private_key, self.github_installation_id)
+        self.token, self.token_expires_at = generate_app_access_token(
+            self.github_app_id, self.github_private_key, self.github_installation_id
+        )
 
         # Check if the token isn't valid.  If not, overwrite it with None
         if not self.is_valid_token():
             if self.logger:
-                self.logger.warning("An app token was generated but could not be validated.")
+                self.logger.warning(
+                    "An app token was generated but could not be validated."
+                )
             self.token = None
             self.token_expires_at = None
 
@@ -226,7 +230,6 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
                 for key, value in env_dict.items()
                 if key.startswith("GITHUB_TOKEN")
             }
-
             if len(env_tokens) > 0:
                 self.logger.info(
                     f"Found {len(env_tokens)} 'GITHUB_TOKEN' environment variables for authentication."
@@ -235,7 +238,9 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
 
         token_managers: List[TokenManager] = []
         for token in personal_tokens:
-            token_manager = PersonalTokenManager(token, rate_limit_buffer=rate_limit_buffer, logger=self.logger)
+            token_manager = PersonalTokenManager(
+                token, rate_limit_buffer=rate_limit_buffer, logger=self.logger
+            )
             if token_manager.is_valid_token():
                 token_managers.append(token_manager)
 
@@ -245,11 +250,15 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
             # "{app_id};;{-----BEGIN RSA PRIVATE KEY-----\n_YOUR_PRIVATE_KEY_\n-----END RSA PRIVATE KEY-----}"
             env_key = env_dict["GITHUB_APP_PRIVATE_KEY"]
             try:
-                app_token_manager = AppTokenManager(env_key, rate_limit_buffer=rate_limit_buffer, logger=self.logger)
+                app_token_manager = AppTokenManager(
+                    env_key, rate_limit_buffer=rate_limit_buffer, logger=self.logger
+                )
                 if app_token_manager.is_valid_token():
                     token_managers.append(app_token_manager)
             except ValueError as e:
-                self.logger.warn(f"An error was thrown while preparing an app token: {e}")
+                self.logger.warn(
+                    f"An error was thrown while preparing an app token: {e}"
+                )
 
         self.logger.info(f"Tap will run with {len(token_managers)} auth tokens")
         return token_managers
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from datetime import datetime, timedelta
 from os import environ
 from random import choice, shuffle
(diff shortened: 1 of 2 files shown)


tap_github/authenticator.py
```diff
@@ -50,6 +52,9 @@ def update_rate_limit(self, response_headers: Any) -> None:
 
     def is_valid_token(self) -> bool:
         """Try making a request with the current token. If the request succeeds return True, else False."""
+        if not self.token:
+            return False
```
TrishGillett, line 56: When working with AppTokenManager, the token may not be generated yet (or may have failed the generation process) at the point at which this function is called, so we need handling that case.

tap_github/authenticator.py
```diff
@@ -135,14 +149,71 @@ def generate_app_access_token(
     if resp.status_code != 201:
         resp.raise_for_status()
 
-    return resp.json()["token"]
+    expires_at = produced_at + timedelta(hours=1)
+    return resp.json()["token"], expires_at
```
TrishGillett, line 153: We'll need to know the expiry for the next PR where refreshing is implemented.

tap_github/authenticator.py
```diff
@@ -135,14 +149,71 @@ def generate_app_access_token(
...
+            self.token = None
+            self.token_expires_at = None
 
 
 class GitHubTokenAuthenticator(APIAuthenticatorBase):
     """Base class for offloading API auth."""
 
+    @staticmethod
+    def get_env():
+        return dict(environ)
```
TrishGillett, line 211: I've had a lot of trouble trying to reliably patch environment variables for testing.  Not sure if there's a better option, but the best I've come up so far is to isolate this into a method so I can easily patch the return value.

tap_github/authenticator.py
```diff
@@ -156,52 +227,40 @@ def prepare_tokens(self) -> List[TokenManager]:
...
-                for key, value in environ.items()
+                for key, value in env_dict.items()
                 if key.startswith("GITHUB_TOKEN")
             }
             if len(env_tokens) > 0:
                 self.logger.info(
                     f"Found {len(env_tokens)} 'GITHUB_TOKEN' environment variables for authentication."
                 )
-                personal_tokens = env_tokens
+                personal_tokens = personal_tokens.union(env_tokens)
```
TrishGillett, line 237: ⚠️ Note: this is a functionality change!  I noticed this edge case:
- `auth_token` is provided in the config
- `additional_auth_tokens` is not provided in the config
- environment variables whose keys start with `GITHUB_TOKEN` are present

The previous logic would have meant that the detected environment variables were used, but `auth_token` wasn't.  I can't imagine we want to ignore a token specified in the config, so I am assuming this is a bug.  To be honest, my preference would be that if anything is specified in the config then all personal tokens in the environment are ignored, but it's possible that could be a potentially breaking change for some users.

(Let me know if you want me to move this change to a different PR, or change this PR's type to `bug` instead of `refactor`. :P)

tap_github/authenticator.py
```diff
@@ -61,7 +66,7 @@ def is_valid_token(self) -> bool:
             return True
         except requests.exceptions.HTTPError:
             msg = (
-                f"A token was dismissed. "
+                f"A token could not be validated. "
```
TrishGillett, line 69: Downgrading this statement since this method now gets called multiple times.  Added another `"A token was dismissed."` log message down below at the point of dismissal (L247), so it'll still appear in the log when tokens are dismissed during initial authentication.

tap_github/authenticator.py
edgarrmondragon: That's indeed a bug 😬. More then happy to communicate this refactor as a bugfix. Thanks for chasing it 🙏.

TrishGillett: Done!

tap_github/tests/test_authenticator.py
```diff
@@ -112,3 +118,309 @@ def test_has_calls_remaining_fails_if_few_calls_remaining_and_reset_time_not_rea
...
+            assert token_manager.github_installation_id == ""
+
+    def test_initialization_with_malformed_env_key(self):
+        with pytest.raises(ValueError) as exc_info:
+            AppTokenManager("12345key\\ncontent")
+
+        assert str(exc_info.value) == (
+            "GITHUB_APP_PRIVATE_KEY could not be parsed. The expected format is "
+            '":app_id:;;-----BEGIN RSA PRIVATE KEY-----\\n_YOUR_P_KEY_\\n-----END RSA PRIVATE KEY-----"'
+        )
```
edgarrmondragon, line 146: Would using [`pytest.raises(ValueError, match=...)`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.raises) be preferred here?

edgarrmondragon (commented):
This looks great @TrishGillett 🎉 . Just a nit about idiomatic pytest.

tap_github/tests/test_authenticator.py
TrishGillett: Good call!  Pushed a change for this. :)

edgarrmondragon (approved):

tap_github/tests/test_authenticator.py
```diff
@@ -112,3 +119,308 @@ def test_has_calls_remaining_fails_if_few_calls_remaining_and_reset_time_not_rea
...
+            assert token_manager.github_app_id == "12345"
+            assert token_manager.github_private_key == "key\ncontent"
+            assert token_manager.github_installation_id == "67890"
+
+    def test_initialization_with_2_part_env_key(self):
+        with patch.object(AppTokenManager, "claim_token", return_value=None):
+            token_manager = AppTokenManager("12345;;key\\ncontent")
+            assert token_manager.github_app_id == "12345"
+            assert token_manager.github_private_key == "key\ncontent"
+            assert token_manager.github_installation_id == ""
```
camerondavison, line 138: I think that https://github.com/MeltanoLabs/tap-github/blob/aade338aaa5fe7a49427e3b7596f9ed45a3bae87/tap_github/authenticator.py#L132 this line is expecting this to be None not `""`

I think we either want to parse this as empty string and then change the other code to look for "falsy" or we want to parse this as "None" and then we do not need to change the code.

I have been getting
```
404 Client Error: Not Found for url: https://api.github.com/app/installations//access_tokens
```
post this change

TrishGillett: Hey @camerondavison, thanks for reporting and sorry this change caused an issue.  Taking a look now. 👀

TrishGillett: Put up a PR to fix this, if you like you can try pegging your version of the tap to the commit `38e1bdf9381c3a3fa17cceada9e1ede7be57ae6b` to see if it fixes your issue.

edgarrmondragon: I've also tagged v1.4.8

PR #289, opened 2024-08-09
feat: App token refreshing

TrishGillett:
This PR implements app token refreshing!  Thank you for your patience with the previous refactoring PRs that prepared for this one.

- As part of `has_calls_remaining`, the amount of time left until estimated token expiry is checked, and tokens are refreshed if expiry is close.
- A new config option `expiry_time_buffer` is added, controlling how many minutes before expiry the app token will be refreshed.  This parallels how `rate_limit_buffer` is used to ensure we don't push tokens all the way to their limits.


TrishGillett:
👋 I ran some large extractions over the weekend using this PR's SHA and a single app key.

The good news is that it ran smoothly for several hours at a time without any authenticator related failure. The longest job I ran took about 8.5 hours, and I can see the `GitHub app token refresh succeeded.` logging at regular intervals.

One slightly confusing find is that whenever the expiry time approached a refresh would happen, and then another refresh would happen say 3-6 minutes later.  Log excerpt demonstrating it:
```
2024-08-10 03:58:53,256 | INFO     | tap-github           | Tap will run with 1 auth tokens
2024-08-10 04:48:55,891 | INFO     | tap-github           | GitHub app token refresh succeeded.
2024-08-10 04:53:55,361 | INFO     | tap-github           | GitHub app token refresh succeeded.
2024-08-10 05:38:55,944 | INFO     | tap-github           | GitHub app token refresh succeeded.
2024-08-10 05:44:31,833 | INFO     | tap-github           | GitHub app token refresh succeeded.
2024-08-10 06:28:56,078 | INFO     | tap-github           | GitHub app token refresh succeeded.
2024-08-10 06:34:59,825 | INFO     | tap-github           | GitHub app token refresh succeeded.
```

I was extracting the issues stream, and when I look at where GitHubTokenAuthenticator is initialized it looks like maybe the parent stream (repos) and child stream (issues) are each constructing and maintaining a GitHubTokenAuthenticator object, meaning there would be two installation access tokens at any given time and each would need to be updated periodically.  Does that match your mental model @edgarrmondragon?  As long as we understand why it's refreshing twice and believe it isn't due to a bug, I don't think this is a problem - multiple installation access tokens can exist for the same app and will just cumulatively contribute to the app's rate limit.

sicarul:
If it helps in anyway, we've been using app tokens with this method without issues since last year, but i haven't refactored the code to address Meltano's feedback: https://github.com/MeltanoLabs/tap-github/pull/201

edgarrmondragon:
> Does that match your mental model @edgarrmondragon?

That makes sense to me. We could always make it singleton in the future, but I don't think it causes issues at the moment?

edgarrmondragon (approved):

TrishGillett:
> We could always make it singleton in the future, but I don't think it causes issues at the moment?

Agreed. 👍

tap_github/tests/test_authenticator.py
```diff
@@ -165,6 +165,135 @@ def test_successful_token_generation(self):
...
+                "X-RateLimit-Reset": "1372700873",
+                "X-RateLimit-Used": "1",
+            }
+
+            token_manager = AppTokenManager("12345;;key\\ncontent;;67890")
+            token_manager.logger = MagicMock()
+            token_manager.token_expires_at = expired_time
+            token_manager.update_rate_limit(mock_response_headers)
+
+            mock_is_valid.return_value = False
```
edgarrmondragon, line 218: Is this needed given the patch already uses `return_value=True`?

edgarrmondragon:
Thanks @TrishGillett, this is a monumental improvement!

PR #291, opened 2024-08-14
fix: fix problem when app token used without installation ID provided

TrishGillett:
Addresses the problem described by @camerondavison [here](https://github.com/MeltanoLabs/tap-github/pull/284#discussion_r1717520222), and introduced in https://github.com/MeltanoLabs/tap-github/pull/284.

If not provided, installation ID should be set to None, since that is the way `generate_app_access_token` expects a missing installation ID to be represented.


edgarrmondragon (approved):

TrishGillett:
This one would be good to ship sooner since it's impacting users, and then I can rebase https://github.com/MeltanoLabs/tap-github/pull/289.

camerondavison:
sorry about the delay, thanks for the quick turn around ! :)

i was able to run with the patched commit and everything did work.

PR #300, opened 2024-08-28
feat: Include closed milestones in milestones stream

TrishGillett:
~⚠️ This may be a disruptive change for some users who are only interested in open milestones and are relying on the fact that this tap only pulls open milestones.  These existing users would need to apply a stream map or downstream filter in order to continue excluding closed milestones.~ Update: we've landed on a way to implement it as optional, with the stream still extracting only open milestones by default.

**Why is it worth making this potentially disruptive change?**
If a user is using this tap to maintain and update a data source about milestones and their states/attributes, but each run only extracts data about open milestones, all past and current milestones will show up as open whether they are open or closed.  Other fields (especially counts of open/closed issues, closed_at, updated_at) would contain potentially stale data corresponding to the last time they were observed before they were closed.  By extracting data about all milestones, previously extracted milestones which are now closed can be updated to a closed state and have any other changed fields updated as well.

The implementation approach for this change follows the pattern used in streams like IssuesStream and PullRequestsStream when diverging from the default query parameters.


edgarrmondragon:
@TrishGillett wdyt about putting behind a (new) setting to make it opt-in?

TrishGillett:
> @TrishGillett wdyt about putting behind a (new) setting to make it opt-in?
I'd be down for that!  Are you thinking we would make an alternative stream similar to how stargazers rest and stargazers graphql coexist?  Or is there a way to add settings to the existing stream?  The latter would be better if that's an option, I don't think I've seen examples of something like that though so a hint or reference would be appreciated!

edgarrmondragon:
> A top level setting like start_date, auth_token, etc, but that would only change behaviour for this stream?

I think this is my preference. Not ideal that a top-level setting affects the behavior of a stream, but there's some precedence in the ecosystem so I think we can live with it.

TrishGillett:
> Not ideal that a top-level setting affects the behavior of a stream, but there's some precedence in the ecosystem so I think we can live with it.

Yeah, it's not my favourite either. I think the least-bad version might be to structure the settings in a way where we won't hate ourselves later if more cases like this come up in other streams and we need to add more.

(Full disclosure: I've been looking into Github's GraphQL API lately and thinking about how there may be reasons to make GraphQL versions of some streams, and I can picture wanting to add more stream options in that case since GraphQL offers a lot more flexibility.  For example, a GraphQL issues stream could support an end date, and hopefully also an option to only get data about issues, instead of the default which includes data about PRs too.)

For a pattern, how would you feel about one of these:
Option 1 (It adds a little bulk to the config but the organization is a little nice IMO)
```
config:
  milestones_stream_params:
    state: all
  other_stream_params:
    thing: value_1
    other_thing: value_2
```

Option 2
```
config:
  milestones_stream_state_param: all
  other_stream_thing_param: value_1
  other_stream_other_thing_param: value_1
```

edgarrmondragon:
> > Not ideal that a top-level setting affects the behavior of a stream, but there's some precedence in the ecosystem so I think we can live with it.
> 
> Yeah, it's not my favourite either. I think the least-bad way to do a top-level setting might be to structure it in a way where we won't hate ourselves later if more cases like this come up in other streams and we need to add more.
> 
> (Full disclosure: I've been looking into Github's GraphQL API lately and thinking about how there may be reasons to make GraphQL versions of some streams, and I can picture wanting to add more stream options in that case since GraphQL offers a lot more flexibility. For example, a GraphQL issues stream could support an end date, and hopefully also an option to only get data about issues, instead of the default which includes data about PRs too.)
> 
> For a pattern, how would you feel about one of these: Option 1 (It adds a little bulk to the config but the organization is a little nice IMO)
> 
> ```
> config:
>   milestones_stream_params:
>     state: all
>   other_stream_params:
>     thing: value_1
>     other_thing: value_2
> ```
> 
> Option 2
> 
> ```
> config:
>   milestones_stream_state_param: all
>   other_stream_thing_param: value_1
>   other_stream_other_thing_param: value_1
> ```

Hey @TrishGillett, sorry for the delay in reviewing. I like Option 1 😄. Wdyt of shorter names, e.g.

```yaml
config:
  milestones:
    state: all
  all_streams:
    thing: value_1
    other_thing: value_2
```

Otherwise I think "nested" is the right approach.

TrishGillett:
Cool!  Will implement. 🙌

TrishGillett:
Pushed an update.  This can be considered fair game for review, although it shouldn't be merged just yet because I still need to do a test run and confirm it works as expected.

tap_github/tap.py
```diff
@@ -83,6 +83,19 @@ def logger(cls) -> logging.Logger:
...
+            th.ObjectType(
+                th.Property(
+                    "state",
+                    th.StringType,
+                    description=(
+                        "Configures which states are of interest. "
+                        "Must be one of [open, closed, all], defaults to open."
+                    ),
+                ),
+            ),
```
edgarrmondragon, line 97: Nit: add a little more metadata to the new settings

```suggestion
            th.ObjectType(
                th.Property(
                    "state",
                    th.StringType,
                    description=(
                        "Configures which states are of interest. "
                        "Must be one of [open, closed, all], defaults to open."
                    ),
                    default="open",
                    allowed_values=["open", "closed", "all"],
                ),
            ),
            descriptions="Setting specific to the 'milestones' stream."
```

meltano.yml
```diff
@@ -35,6 +35,8 @@ plugins:
       kind: array
     - name: user_ids
       kind: array
+    - name: milestones
+      kind: object
```
edgarrmondragon, line 39: Nit: expand the nested setting

```suggestion
    - name: milestones.state
      kind: options
      options:
      - label: Open
        value: open
      - label: Closed
        value: closed
      - label: all
        value: All
```

tap_github/tap.py
TrishGillett: Nice!

tap_github/repository_streams.py
```diff
@@ -635,6 +635,18 @@ class MilestonesStream(GitHubRestStream):
...
 
+    def get_url_params(
+        self, context: dict | None, next_page_token: Any | None
+    ) -> dict[str, Any]:
+        """Return a dictionary of values to be used in URL parameterization."""
+        assert context is not None, f"Context cannot be empty for '{self.name}' stream."
+        params = super().get_url_params(context, next_page_token)
+
+        config_params = self.config.get("milestones") or {}
+        params["state"] = config_params.get("state") or "open"
```
TrishGillett, line 646: @edgarrmondragon Now that we've set a default value in tap.py, does that mean I no longer need to worry about the case where `self.config["milestones"]` or `self.config["milestones"]["state"]` aren't provided, and don't need to reassert the default here?

TrishGillett:
Done test runs:
- ✅ When not using the new config, you still get only open milestones
- ✅ When using the new config with state = `all`, you get all milestones
🎉

edgarrmondragon (approved):

tap_github/repository_streams.py
edgarrmondragon: I think it's ok to check for `milestones`, but if that is present I believe it's safe to assume the JSON schema validator will inject the `milestones.state` default.

TrishGillett: Apologies for the delay.  I've pushed a refactor, and am hoping to test it in the next couple days. 🙏

edgarrmondragon: @TrishGillett the changes LGTM, so do let me know if are looking good on the integration side 😄

TrishGillett: Hey!  Here's how my tests went for the current implementation:
- ✅ Running with the milestones state param set to `all` gives x issues in closed state and y issues in open state
- ✅ Running with the milestones state param set to `open` gives y issues in open state
- ✅ Running with the milestones state param set to `closed` gives x issues in closed state
- ✅ Running with the milestones state param set to `blah` produces an error: `"singer_sdk.exceptions.ConfigValidationError: Config validation failed: 'blah' is not one of ['open', 'closed', 'all']"`
- ✅ Running with the milestones state param not set gives y issues in open state (defaulting to `open` behaviour as expected)
- ❓ I hoped that adding other params nested under `milestones` that are not described in meltano.yml or tap.py would produce an error, but it did not error and completed a full run. Specifically this is the wrong config I tested:
```
        milestones:
          states: [open, closed]
```
 Not sure how we feel about that.  I tried adding [this line here](https://github.com/MeltanoLabs/tap-github/pull/300/files#diff-336fc7fcba52f0e447a92c81a8467e15e394def66eb320092b1cafcb06455376R99) to see if it would reject `states`, but no change.
 
 So, if people write the wrong nested param it won't complain, and if they happen to write a nested param that is [valid for the endpoint](https://docs.github.com/en/rest/issues/milestones?apiVersion=2022-11-28) I'm assuming it would actually be passed through because I'm updating with [all provided params](https://github.com/MeltanoLabs/tap-github/pull/300/files#diff-06dc9c6115cbc069ce355913de0c101fedf6956d6f6b4873c5112434596934d3R647). How do you feel about that?

TrishGillett: If we're uneasy about that last case, I can hardcode the list of params to check for rather than just taking whatever params are provided.  Let me know!

edgarrmondragon: I think `additional_properties` is working as expected. Using your bad config example, I get

```console
$ meltano config tap-github test
Plugin configuration is invalid
JSONSchema was: {'type': 'object', 'properties': {'user_agent': {'type': ['string', 'null']}, 'metrics_log_level': {'type': ['string', 'null']}, 'auth_token': {'type': ['string', 'null'], 'description': 'GitHub token to authenticate with.'}, 'additional_auth_tokens': {'type': ['array', 'null'], 'items': {'type': ['string']}, 'description': 'List of GitHub tokens to authenticate with. Streams will loop through them when hitting rate limits.'}, 'auth_app_keys': {'type': ['array', 'null'], 'items': {'type': ['string']}, 'description': 'List of GitHub App credentials to authenticate with. Each credential can be constructed by combining an App ID and App private key into the format `:app_id:;;-----BEGIN RSA PRIVATE KEY-----\n_YOUR_P_KEY_\n-----END RSA PRIVATE KEY-----`.'}, 'rate_limit_buffer': {'type': ['integer', 'null'], 'description': 'Add a buffer to avoid consuming all query points for the token at hand. Defaults to 1000.'}, 'expiry_time_buffer': {'type': ['integer', 'null'], 'description': ('When authenticating as a GitHub App, this buffer controls how many minutes before expiry the GitHub app tokens will be refreshed. Defaults to 10 minutes.',)}, 'searches': {'type': ['array', 'null'], 'items': {'type': 'object', 'properties': {'name': {'type': ['string']}, 'query': {'type': ['string']}}, 'required': ['name', 'query']}}, 'organizations': {'type': ['array', 'null'], 'items': {'type': ['string']}}, 'repositories': {'type': ['array', 'null'], 'items': {'type': ['string']}}, 'user_usernames': {'type': ['array', 'null'], 'items': {'type': ['string']}}, 'user_ids': {'type': ['array', 'null'], 'items': {'type': ['string']}}, 'milestones': {'type': ['object', 'null'], 'properties': {'state': {'type': ['string', 'null'], 'default': 'open', 'description': 'Configures which states are of interest. Must be one of [open, closed, all], defaults to open.', 'enum': ['open', 'closed', 'all']}}, 'additionalProperties': False, 'description': "Setting specific to the 'milestones' stream."}, 'start_date': {'type': ['string', 'null'], 'format': 'date-time'}, 'stream_maps': {'type': ['object', 'null'], 'properties': {}}, 'stream_map_config': {'type': ['object', 'null'], 'properties': {}}, 'skip_parent_streams': {'type': ['boolean', 'null'], 'description': 'Set to true to skip API calls for the parent streams (such as repositories) if it is not selected but children are'}, 'flattening_enabled': {'type': ['boolean', 'null'], 'description': "'True' to enable schema flattening and automatically expand nested properties."}, 'flattening_max_depth': {'type': ['integer', 'null'], 'description': 'The max depth to flatten schemas.'}, 'batch_config': {'type': ['object', 'null'], 'properties': {'encoding': {'type': ['object', 'null'], 'properties': {'format': {'type': ['string', 'null'], 'description': 'Format to use for batch files.', 'enum': ['jsonl']}, 'compression': {'type': ['string', 'null'], 'description': 'Compression format to use for batch files.', 'enum': ['gzip', 'none']}}, 'description': 'Specifies the format and compression of the batch files.'}, 'storage': {'type': ['object', 'null'], 'properties': {'root': {'type': ['string', 'null'], 'description': 'Root path to use when writing batch files.'}, 'prefix': {'type': ['string', 'null'], 'description': 'Prefix to use when writing batch files.'}}, 'description': 'Defines the storage layer to use when writing batch files'}}}}}
```

whereas if I use the correct setting, I get

```console
$ meltano config tap-github test
Plugin configuration is invalid
RuntimeError: All GitHub tokens have hit their rate limit. Stopping here.
```

---

That said, I got another concern after interacting with the extractor settings via `meltano config ... set`. There is a `repositories` setting, but unlike `milestones`, it doesn't contain stream-specific settings but a list of repos to extract from. I wonder if we should nest `milestones` under `streams.milestones` instead.

TrishGillett: That's a good point about `repositories` being a pre-existing config that breaks the pattern. I'm okay with nesting under something as a solution.  I'll go with whatever you feel best about, but if I can put another option out there, how would you feel about `stream_options`?  I think the generality of the name `streams` makes it a little unintuitive when you picture all the other stream related config that's under other names like `select`, `stream_maps`, and `stream_map_config`.

edgarrmondragon: I like `stream_options`

TrishGillett: Updated and ran some test runs.  The test runs look good:
- ✅ Running with the milestones state param set to `all` gives x issues in closed state and y issues in open state
- ✅ Running with the milestones state param set to `open` gives y issues in open state
- ✅ Running with the milestones state param set to `closed` gives x issues in closed state
- ✅ Running with the milestones state param set to `blah` produces an error: `"singer_sdk.exceptions.ConfigValidationError: Config validation failed: 'blah' is not one of ['open', 'closed', 'all']"`
- ✅ Running with the milestones state param not set gives y issues in open state (defaulting to `open` behaviour as expected)
- ✅  With wrongly structured params such as
```
  stream_options:
    milestones:
      states: [open, closed]
```
 an error is thrown: `ConfigValidationError: Config validation failed: Additional properties are not allowed ('states' was unexpected)`

edgarrmondragon (approved):
LGTM!

Thanks for your patience @TrishGillett 😄

PR #302, opened 2024-09-08
Enable running with multiple app keys provided via config

TrishGillett:
Adds a way to input the keys for multiple github apps, which people might want to do in order to increase capacity and protect against rate limits.
As a bonus, this also gives users the option to provide their app key via config rather than having it inferred, and the option to use environment variable names of their choice.


meltano.yml
```diff
@@ -19,6 +19,8 @@ plugins:
       kind: password
     - name: additional_auth_tokens
       kind: array
+    - name: auth_app_keys
```
TrishGillett, line 22: For simplicity, I'm opting to only create an `auth_app_keys` option and not mirror the `auth_token` + `additional_auth_tokens` structure.

edgarrmondragon: Can we add this to https://github.com/MeltanoLabs/tap-github/blob/53057a2388bb70c0919ce53c017106f8bdb174f7/tap_github/tap.py#L36 as well?

There's some unfortunate amount of duplication between that and `meltano.yml` at the moment.

TrishGillett: Will update later tonight or tomorrow, thanks for catching.  I'll add `expiry_time_buffer` too while I'm add it, I missed adding an entry for it in https://github.com/MeltanoLabs/tap-github/pull/289.

edgarrmondragon: Thanks!

PR #357, opened 2025-02-13
feat: Add GitHub Streams (branches, tags)

malexmad:
Adding 2 new github streams 
- [BranchesStream](https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#list-branches)
- [TagsStream](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-tags)


edgarrmondragon (approved):
Thanks @malexmad!

PR #345, opened 2025-01-23
feat: add `PullRequestDiffsStream` as child to `PullRequestsStream`

angelachenn:
This PR adds a child stream for pull request diffs to the parent pull requests stream. References [Parent-Child Streams](https://gitlab.com/meltano/sdk/-/blob/main/docs/parent_streams.md) and Github REST API documentation.

- also adds `pull_id` to the PR stream parent context, making `{org, repo, repo_id, pull_number, pull_id}`. Gives users more flexibility in using or joining their tables.
- I followed code patterns that I noticed, except for matching error status codes against their respective messages; this is more to avoid a dependency on Github error messages. If they were to change them for whatever reason, that'd be a breaking change.

Some clarity on the error catching: all of them are related to the diff size being too large for to process and/or stream.
- `406`, `422` and `502` are all from the Github REST API's end. 
  - `406` being `diff too large` either because it exceeds the max # of files (300) or lines changed (20000)
  - `422` being server error because `diff is taking too long to generate`
  - `502` Server Error
- PR diffs that are small enough to call via the API but still quite large will result in this error:
Screenshot 2025-01-23 at 4 21 08 PM
Modifying the buffer size is risky and doesn't really prevent related issues from resurfacing downstream. Luckily intervening at the parsing step worked, I've set the content-size limit to 40MiB to allow for some overhead. 

PRs that cause those errors are skipped. I've tested with a pipeline on my end and it all seems to be working successfully.
```
2025-01-23 08:16:29,000 | INFO     | tap-github.pull_request_diffs | Skipping PR due to 422 error: Server Error: Sorry, this diff is taking too long to generate.
2025-01-23 08:17:59,294 | INFO     | tap-github.pull_request_diffs | Skipping PR due to 406 error: Sorry, the diff exceeded the maximum number of files (300). Consider using 'List pull requests files' API or locally cloning the repository instead.
2025-01-23 08:43:26,939 | INFO     | tap-github.pull_request_diffs | Skipping PR. The diff size (55.72 MiB) exceeded the maximum size limit of 40 MiB.
```
My initial brute-force approach was just to shove another call to the diff in the `post_process` method within `PullRequestsStreams` based on a config setting, but that results in getting instantly rate-limited lol. 

Please advise if I missed anything. Thanks!

PR #359, opened 2025-02-18
feat: include records for failed PR diff fetches in `PullRequestDiffsStream`

angelachenn:
This PR also contains some forward fixes following https://github.com/MeltanoLabs/tap-github/pull/345, namely
- a mis-indented `return` statement was causing all PR diff records to be 'skipped'
- added `404` to list of known Github API errors

For completeness purposes, 2 new PR diff properties have been added: `success` and `error_message`.
- If a PR diff is successfully fetched, `success`=`True` and `error_message` is null.
- If a PR diff is 'skipped' for any reason (e.g., GIthub API error, diff size is too large) a record will still be made for it with `success`=`False` and the respective `error_message` (so users can at least know which PRs have diffs that weren't retrieved).

 I've tested my changes on my end and it all seems to be working successfully (records being produced with the new attributes set correctly), though please advise if I missed anything. Thank you!

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1449,6 +1449,7 @@ class PullRequestDiffsStream(GitHubRestStream):
     parent_stream_type = PullRequestsStream
     ignore_parent_replication_key = False
     state_partitioning_keys: ClassVar[list[str]] = ["repo", "org"]
+    # Known Github API errors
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):

edgarrmondragon:
Thanks!

PR #361, opened 2025-02-19
feat: Updated primary key to include name in `branches` and `tags` streams

malexmad:
Fixing a bug in the new BranchesStream and TagsStream that was brought in this last [PR](https://github.com/MeltanoLabs/tap-github/pull/357). By not having the `name` in the primary keys it returns one data entry per org/repo. This should allow the stream to pick up all tags and branches in a given repo now. (tested this downstream and it works as expected)


edgarrmondragon (approved):
Thanks @malexmad!

PR #363, opened 2025-02-20
feat: Added `deployments` and `deployment_statuses` streams

malexmad:
Adding 2 new stream for deployments:
- [deployment list](https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#list-deployments)
- [deployment status list](https://docs.github.com/en/rest/deployments/statuses?apiVersion=2022-11-28#list-deployment-statuses)


tap_github/repository_streams.py
```diff
@@ -2625,3 +2625,131 @@ class TagsStream(GitHubRestStream):
...
         th.Property("node_id", th.StringType),
     ).to_dict()
+
+
+class DeploymentsStream(GitHubRestStream):
+    """A stream dedicated to fetching deployments in a repository."""
+
+    name = "deployments"
+    path = "/repos/{org}/{repo}/deployments"
+    primary_keys: ClassVar[list[str]] = ["repo", "org", "node_id"]
```
edgarrmondragon, line 2635: Shouldn't `node_id` be sufficient to globally and uniquely identify a deployment?

```suggestion
    primary_keys: ClassVar[list[str]] = ["node_id"]
```

https://docs.github.com/en/graphql/guides/using-global-node-ids

tap_github/repository_streams.py
```diff
@@ -2625,3 +2625,131 @@ class TagsStream(GitHubRestStream):
...
+            "repo_id": context["repo_id"] if context else None,
+        }
+
+
+class DeploymentStatusesStream(GitHubRestStream):
+    """A stream dedicated to fetching deployment statuses in a repository."""
+
+    name = "deployment_statuses"
+    path = "/repos/{org}/{repo}/deployments/{deployment_id}/statuses"
+    primary_keys: ClassVar[list[str]] = ["repo", "org", "deployment_id"]
```
edgarrmondragon, line 2706: Similarly, I think we could away with

```suggestion
    primary_keys: ClassVar[list[str]] = ["node_id"]
```

tap_github/repository_streams.py
malexmad: yep updated both suggestion

edgarrmondragon (approved):

PR #370, opened 2025-03-04
fix: check for null row["issue"] in issue events stream

TrishGillett:
We ran into an `AttributeError: 'NoneType' object has no attribute 'pop'` error while testing the IssueEventsStream, it's originating from [this line](https://github.com/MeltanoLabs/tap-github/blob/9033fe4d2b28386c6e8c6fb02262d80b862cd05d/tap_github/repository_streams.py#L1032).

It appears to be possible for row to have an `issue` key but for its value in the dict to be None.  This PR adds a check for it.

We've tested the fix on our end successfully.

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -1028,9 +1028,9 @@ class IssueEventsStream(GitHubRestStream):
 
     def post_process(self, row: dict, context: dict | None = None) -> dict:
         row = super().post_process(row, context)
-        if "issue" in row and row["issue"] is not None:
-            row["issue_number"] = int(row["issue"].pop("number"))
-            row["issue_url"] = row["issue"].pop("url")
+        if issue := row.get("issue"):
+            row["issue_number"] = int(issue.pop("number"))
+            row["issue_url"] = issue.pop("url")
         else:
             self.logger.debug(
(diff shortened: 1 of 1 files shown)


tap_github/repository_streams.py
```diff
@@ -1028,7 +1028,7 @@ def get_records(self, context: dict | None = None) -> Iterable[dict[str, Any]]:
 
     def post_process(self, row: dict, context: dict | None = None) -> dict:
         row = super().post_process(row, context)
-        if "issue" in row:
+        if "issue" in row and row["issue"] is not None:
             row["issue_number"] = int(row["issue"].pop("number"))
             row["issue_url"] = row["issue"].pop("url")
```
edgarrmondragon, line 1033:
```suggestion
        if issue := row.get("issue"):
            row["issue_number"] = int(issue.pop("number"))
            row["issue_url"] = issue.pop("url")
```

edgarrmondragon:
Thanks @TrishGillett!

tap_github/repository_streams.py
TrishGillett: LGTM!

edgarrmondragon (approved):

Issue #149, opened 2022-06-22
State bookmarks should use ids instead of repo names


laurentS:
Following the same logic as what is described in #110 and the linked PR, the state bookmarks should use stable IDs for their context, as opposed to the current `org / repo`:
If a repo is renamed between 2 runs of the tap, the later run would ignore the bookmark, as the repo name would not match anymore. While not a disaster, it would essentially force a full resync of the history of the stream, which is a waste of time/energy/api quotas.

Example of a state bookmark right now:
```json
    {
      "context": { "org": "scipy", "repo": "scipy" },  // this should use a stable id
      "replication_key": "updated_at",
      "replication_key_value": "2022-05-06T14:29:18Z",
      "replication_key_signpost": "2022-06-22T09:07:14.047100+00:00",
      "starting_replication_value": "2022-05-06T14:29:18Z",
      "progress_markers": {
        "Note": "Progress is not resumable if interrupted.",
        "replication_key": "updated_at",
        "replication_key_value": "2022-06-22T07:55:32Z"
      }
    },
```

laurentS:
Relates to #122

PR #376, opened 2025-03-21
Create Discussions Streams

JeannC:


tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                  ... on User{
+                    node_id: id
+                    id: databaseId
+                    login
+                    avatar_url: avatarUrl
+                    html_url: url
+                    type: __typename
+                    site_admin: isSiteAdmin
+                    }
+                  }
```
TrishGillett, line 1973: I notice in our test runs this is returning nulls for all author fields when the Actor is a Bot or other non-User type.  Everywhere we're doing this `... on User{`, can we try doing this 👇  instead and see A) it still runs and B) if these columns get filled in properly for non-Users?

```
... on Actor {
  node_id: id
  id: databaseId
  login
  avatar_url: avatarUrl
  html_url: url
  type: __typename
}
... on User {
  site_admin: isSiteAdmin
}
```

The intent is that it will try to get those 6 fields for all Actor types, and only try to get site_admin if the actor is also a User.

Note this is making an assumption that if any new Actor implementations are created they will have all 6 of these fields.  All current implementations have all 6 fields, but [Actor](https://docs.github.com/en/graphql/reference/interfaces#actor) only requires 4 of them, the missing two being id and databaseId. So, if a new Actor implementation is ever created AND they choose not to make both `id` and `databaseId` available then this could break and need an update.  I think the alternative would be to do something like:
```
... on Actor {
  login
  avatar_url: avatarUrl
  html_url: url
  type: __typename
}
... on User {
  node_id: id
  id: databaseId
  site_admin: isSiteAdmin
}
... on Bot {
  node_id: id
  id: databaseId
}
... on Organization {
  node_id: id
  id: databaseId
}
... on Mannequin {
  node_id: id
  id: databaseId
}
... on EnterpriseUserAccount {
  node_id: id
  id: databaseId
}
```
which would be safe for all current and future types, but is a lot to be copy-pasting everywhere actors appear in all three of these streams.

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+        params = super().get_url_params(context, next_page_token)
+        self.logger.info(f"URL Params: {params}")
+        self.logger.info(f"Context: {context}")
+        self.logger.info(f"State: {self.get_starting_replication_key_value(context)}")
+        return params
+
+    @property
+    def query(self) -> str:
+        """Return dynamic GraphQL query."""
+        # Graphql id is equivalent to REST node_id. To keep the tap consistent, we rename "id" to "node_id".  # noqa: E501
```
TrishGillett, line 1937: May as well add this to the docstring:
```suggestion
        """
        Return dynamic GraphQL query.
        Note: To keep the tap consistent, we rename id to node_id and databaseId to id.
        """
```
or however the linter wants to style it.
(and same for the other streams)

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                  created_at: createdAt
+                  published_at: publishedAt
+                  last_edited_at: lastEditedAt
+                  answer_chosen_at: answerChosenAt
+                  updated_at: updatedAt
+                  closed_at: closedAt
+
+                  created_via_email: createdViaEmail
+
+                  author {
```
TrishGillett, line 1963: Extra spaces:
```suggestion
                  created_via_email: createdViaEmail
                  author {
```

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                    }
+                  }
+                  author_association: authorAssociation
+                  category {
+                    node_id: id
+                    slug
+                    name
+                    description
+                  }
+                  labels(first: 100) {
```
TrishGillett, line 1981: Can you please add something to the class docstring to state that at most 100 labels and 100 reactions will be extracted?

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                    edges {
+                      node {
+                        node_id: id
+                        created_at: createdAt
+                        name
+                        description
+                        url
+                        resource_path: resourcePath
+                        color
+                        default: isDefault
```
TrishGillett, line 1991: Looks like there are more fields available for labels when using graphQL as opposed to REST.  That's neat.  I think it's okay to not limit this to just the fields available in REST. 👍

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+    def get_url_params(
+        self,
+        context: dict | None,
+        next_page_token: Any | None,  # noqa: ANN401
+    ) -> dict[str, Any]:
+        """Return a dictionary of values to be used in URL parameterization."""
+        params = super().get_url_params(context, next_page_token)
+        self.logger.info(f"URL Params: {params}")
+        self.logger.info(f"Context: {context}")
+        self.logger.info(f"State: {self.get_starting_replication_key_value(context)}")
```
TrishGillett, line 1931: We can remove this logging, right? (and same for the other streams)

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+        return params
+
+    @property
+    def query(self) -> str:
+        """Return dynamic GraphQL query."""
+        # Graphql id is equivalent to REST node_id. To keep the tap consistent, we rename "id" to "node_id".  # noqa: E501
+
+        return """
+            query DiscussionCategories($repo: String!, $org: String!, $nextPageCursor_0: String) {
+            repository(name: $repo, owner: $org) {
```
TrishGillett, line 2150: Indentation got mangled when pasting the query in, I think.

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                discussionCategories(first: 100, after: $nextPageCursor_0) {
+                pageInfo {
+                  hasNextPage_0: hasNextPage
+                  startCursor_0: startCursor
+                  endCursor_0: endCursor
+                }
+                nodes {
+                node_id: id
```
TrishGillett, line 2158: To confirm: there's no databaseId counterpart for the discussion category object?

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+        self.logger.info(f"State: {self.get_starting_replication_key_value(context)}")
+        return params
+
+    @property
+    def query(self) -> str:
+        """Return dynamic GraphQL query."""
+        # Graphql id is equivalent to REST node_id. To keep the tap consistent, we rename "id" to "node_id" and "databaseId" to "id".  # noqa: E501
+        return """
+        query DiscussionsComments($repo: String!, $org: String!, $nextPageCursor_0: String, $nextPageCursor_1: String) {
+        repository(name: $repo, owner: $org) {
```
TrishGillett, line 2219: More indentation needing a touch up.  Also, the other two queries used 2 space indents and this one uses 4 spaces.  Can we stick with whichever is more common in the tap's other graphql streams?

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                        node_id: id
+                        id: databaseId
+                        login
+                        avatar_url: avatarUrl
+                        html_url: url
+                        type: __typename
+                        site_admin: isSiteAdmin
+                    }
+                    }
+                    replies(first: 10) {
```
TrishGillett, line 2279: Is the 10 just for testing or should it ship like that?

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                    nodes {
+                        comment_id: id
+                        replyTo {
+                        replyTo_id: id
```
TrishGillett, line 2283: Typo:
```suggestion
                        reply_to_id: id
```
Not sure if we're able to rename the line above as well (is `reply_to: replyTo {` valid syntax?).

Remind me again - what is the relationship between the comment_id and the reply_to_id? If either one is just repeating the id/node_id from [up above](https://github.com/MeltanoLabs/tap-github/pull/376/files#diff-06dc9c6115cbc069ce355913de0c101fedf6956d6f6b4873c5112434596934d3R2234) we can skip it IMO.

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                        }
+                    }
+                    }
+                    reactions(first: 10) {
```
TrishGillett, line 2287: Same question about whether 10 is what we mean to ship with or whether it should be 100.

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+                    }
+                  }
+                  upvote_count: upvoteCount
+                  reactions(first: 100) {
+                    edges {
+                      node {
+                        reaction_type: content
+                        reacted_at: createdAt
+                        user {
+                          login
```
TrishGillett, line 2034: I wonder if we should grab the same user fields we grab everywhere else, for tap consistency. 🤔

tap_github/repository_streams.py
```diff
@@ -1907,6 +1907,457 @@ def query(self) -> str:
...
+            th.Property("name", th.StringType),
+            th.Property("description", th.StringType),
+            th.Property("url", th.StringType),
+            th.Property("resource_path", th.StringType),
+            th.Property("color", th.StringType),
+            th.Property("default", th.BooleanType),
+        )
+    )
+
+    reactions_array = th.ArrayType(
```
TrishGillett, line 2077: Since we will use this one twice (here and discussion comments), we could refactor it out to the `schema_objects.py` file.  Just would need to name it something distinct from the existing `reactions_object` there.

tap_github/repository_streams.py
```diff
@@ -1907,6 +1909,293 @@ def query(self) -> str:
...
 
+class DiscussionsStream(GitHubGraphqlStream):
+    """Defines stream fetching discussions from each repository."""
+
+    name = "discussions"
+    query_jsonpath = "$.data.repository.discussions.nodes.[*]"
+    primary_keys: ClassVar[list[str]] = ["id"]
+    replication_key = "updated_at"
+    parent_stream_type = RepositoryStream
+    state_partitioning_keys: ClassVar[list[str]] = ["repo", "org"]
```
TrishGillett, line 1920: It looks like there's a best practice of using `repo_id` here instead of the repo and org names ([ref 1](https://github.com/MeltanoLabs/tap-github/issues/122), [ref 2](https://github.com/MeltanoLabs/tap-github/issues/149))!

PR #379, opened 2025-03-25
Filter replication key items for `use_fake_since_parameter`

cbrammer:
This tap uses a custom `use_fake_since_parameter` for items API's in GitHub that don't have a `since` parameter. This effects all items that use `use_fake_since_parameter`. This filters out any returned items that are before the `start_date`.

An example is PullRequestStream. It would get an entire page of 100 items, and 99 of them could be past the `since` date. But all 100 were then spinning up child streams to get comments/commits etc. That was causing a huge extra usage of the API request limit. This filters those out.

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -326,6 +326,52 @@ class GitHubRestStream(RESTStream):
         return {"rest": 1, "graphql": 0, "search": 0}
 
 
+class GitHubDiffStream(GitHubRestStream):
+    """Base class for GitHub diff streams."""
+
+    @property
+    def http_headers(self) -> dict:
+        """Return the http headers needed for diff requests."""
+        headers = super().http_headers
+        headers["Accept"] = "application/vnd.github.v3.diff"
+        return headers
+
+    def parse_response(self, response: requests.Response) -> Iterable[dict]:
+        """Parse the response to yield the diff text instead of an object
+        and prevent buffer overflow."""
+        if response.status_code != 200:
+            contents = response.json()
+            self.logger.info(
+                "Skipping %s due to %d error: %s",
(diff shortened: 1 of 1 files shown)


tap_github/client.py
```diff
@@ -142,6 +145,9 @@ def get_url_params(
         context: dict | None,
         next_page_token: Any | None,  # noqa: ANN401
     ) -> dict[str, Any]:
+        # save the context from the requests so it can be available to the parse_response method
+        self.context = context
```
edgarrmondragon, line 149: AFAIK this is not necessary. The stream class already has a `context` attribute.

cbrammer: Unfortunately, I don't think it is available 😞 on the core `RESTStream` class. That is why all other method signatures include it to be passed in. For some reason it was excluded from this method.

tap_github/client.py
```diff
@@ -57,6 +57,9 @@ def url_base(self) -> str:
     replication_key: str | None = None
     tolerated_http_errors: ClassVar[list[int]] = []
 
+    # Save the context from the requests so it can be available to the parse_response method
+    context: dict | None = None
+
```
edgarrmondragon, line 62:
```suggestion
```

tap_github/client.py
```diff
@@ -142,6 +145,9 @@ def get_url_params(
         context: dict | None,
         next_page_token: Any | None,  # noqa: ANN401
     ) -> dict[str, Any]:
+        # save the context from the requests so it can be available to the parse_response method
+        self.context = context
+
```
edgarrmondragon, line 150:
```suggestion
```

edgarrmondragon (changes requested):

PR #372, opened 2025-03-06
feat: add `CommitDiffsStream`, `PullRequestCommitDiffsStream`, refactor diff logic into `GitHubDiffStream`

angelachenn:
This PR adds child streams for fetching diffs of the respective parent streams. Referenced [Parent-Child Streams](https://gitlab.com/meltano/sdk/-/blob/main/docs/parent_streams.md) and followed the same format as changes in #345.
- `CommitDiffsStream` -> `CommitsStream` as parent
- `PullRequestCommitDiffsStream` -> `PullRequestCommitsStream` as parent
- renamed `PullRequestCommits` to `PullRequestCommitsStream` to match convention

Like #345, parent context is also added to relevant streams to give users more flexibility in using or joining their tables. 
I've tested changes on my end successfully and as always, please advise if I missed anything. Thank you!

Latest commit refactors the repetition between the 3 diff streams into a reusable block (`GithubDiffStream`)

PR #380, opened 2025-03-27
fix: Move tolerated errors to parent `GitHubDiffStream` class

angelachenn:
Small followup to https://github.com/MeltanoLabs/tap-github/pull/372. 
Realized that the `tolerated_http_errors` should be the same for all `GithubDiffsStream` class members, not only `PullRequestDiffs`. I have verified that changes work as expected 👍🏻 

Thank you in advance!


edgarrmondragon (approved):
Thanks @angelachenn!

PR #407, opened 2025-05-29
feat: Add contributors stream

TrishGillett:
Trying to pick up the torch from #268, since we also have a need for this stream. :)

Added @MindaugasN to the commit for credit.


TrishGillett:
(Eek, I just realized I typoed the stream name in the commit message. No issue in the code though.  Apologies for any confusion!)

PR #408, opened 2025-05-30
fix: add silent fails for collaborators + contributors

ondramie:
hi, adding some acceptable errors:
1. some repos require `admin:org` and i don't want to give `admin:write` scope. 
2. some repos don't have those paths


edgarrmondragon (approved):
Thanks!

PR #409, opened 2025-05-31
feat(taps): Add streams for projectv2, project v2 field configurations and project v2 item field values

rrxxww:
Add 3 streams to pull corresponding Github ProjectV2 data:
- ProjectsStream for [ProjectV2](https://docs.github.com/en/graphql/reference/objects#projectv2) objects
- ProjectFieldConfigurationsStream for [ProjectFieldConfiguration](https://docs.github.com/en/graphql/reference/objects#projectv2fieldconfiguration) objects
- ProjectItemFieldValuesStream for [ProjectItemFieldValue](https://docs.github.com/en/graphql/reference/unions#projectv2itemfieldvalue) objects

Question to Meltano team: what fields should be included in the stream and what should be excluded? Current implementation tries to include as many fields as possible available from Graphql.

### DIFF
--- a/tap_github/organization_streams.py
+++ b/tap_github/organization_streams.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import hashlib
 from typing import TYPE_CHECKING, Any, ClassVar
 
-import requests
 from singer_sdk import typing as th  # JSON Schema typing helpers
 from singer_sdk.exceptions import FatalAPIError
 
--- a/tap_github/organization_streams.py
+++ b/tap_github/organization_streams.py
@@ -606,7 +606,11 @@ class ProjectItemsStream(GitHubGraphqlStream):
     }
 
     def request_records(self, context: Context | None) -> Iterable[dict]:
-        """Request records from the API, handling FORBIDDEN errors gracefully."""
+        """Request records from the API, handling FORBIDDEN errors gracefully.
+        
+        TODO: should rewrite to use validate_response once
+        https://github.com/meltano/sdk/issues/280 is implemented.
+        """
         try:
             yield from super().request_records(context)
         except FatalAPIError as e:
--- a/tap_github/organization_streams.py
+++ b/tap_github/organization_streams.py
@@ -607,7 +607,7 @@ class ProjectItemsStream(GitHubGraphqlStream):
 
     def request_records(self, context: Context | None) -> Iterable[dict]:
         """Request records from the API, handling FORBIDDEN errors gracefully.
-        
+
         TODO: should rewrite to use validate_response once
         https://github.com/meltano/sdk/issues/280 is implemented.
         """
--- a/tap_github/organization_streams.py
+++ b/tap_github/organization_streams.py
@@ -2,11 +2,14 @@
 
 from __future__ import annotations
 
+import hashlib
 from typing import TYPE_CHECKING, Any, ClassVar
 
+import requests
 from singer_sdk import typing as th  # JSON Schema typing helpers
+from singer_sdk.exceptions import FatalAPIError
 
-from tap_github.client import GitHubRestStream
+from tap_github.client import GitHubGraphqlStream, GitHubRestStream
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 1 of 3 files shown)


tap_github/organization_streams.py
```diff
@@ -211,3 +213,732 @@ class TeamRolesStream(GitHubRestStream):
...
+        if not context:
+            return row
+
+        # Add context fields
+        row["org"] = context["org"]
+        row["project_number"] = context["project_number"]
+
+        # Process dynamic field values into a list of objects
+        field_values_output: list[dict[str, Any]] = []
+        for field_config in self._current_project_field_configurations:
```
TrishGillett, line 804: Github includes Title and Status in projects by default when you first create them, and most projects use them them. In the projects we tested, they make up almost two thirds of the field values used.  Can we pop these two fields out into dedicated columns `title` and `state` (nullable for rows where the fields don't exist), and have `field_values` contain just the remaining fields?

tap_github/organization_streams.py
```diff
@@ -211,3 +213,732 @@ class TeamRolesStream(GitHubRestStream):
...
+                        ),
+                        required=False,  # Only present for ProjectV2SingleSelectField
+                    ),
+                )
+            ),
+        ),
+    ).to_dict()
+
+
+class ProjectItemFieldValuesStream(GitHubGraphqlStream):
```
TrishGillett, line 560: I vote for calling this one simply ProjectItemsStream. :)

tap_github/tap.py
```diff
@@ -144,7 +144,7 @@ def discover_streams(self) -> list[Stream]:
         ):
             raise ValueError(
                 "This tap requires one and only one of the following path options: "
-                f"{Streams.all_valid_queries()}."
+                f"{Streams.all_valid_queries()}, provided config: {self.config}"
```
TrishGillett, line 147: 👍  Nice improvement for debugging config issues :)

tap_github/organization_streams.py
```diff
@@ -211,3 +213,732 @@ class TeamRolesStream(GitHubRestStream):
...
+                  ... on ProjectV2IterationField {
+                    id: databaseId
+                    node_id: id
+                    name
+                    data_type: dataType
+                    created_at: createdAt
+                    updated_at: updatedAt
+                    configuration {
+                      duration
+                      start_day: startDay
```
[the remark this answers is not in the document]
rrxxww: Actually this is a Integer field [The iteration's start day of the week.](https://docs.github.com/en/graphql/reference/objects#projectv2fieldconfiguration:~:text=The%20iteration%27s%20start%20day%20of%20the%20week. ) so I think it is better stay as `start_day`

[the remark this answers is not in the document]
TrishGillett: Ahhh, good to know. Sounds good!

tap_github/organization_streams.py
TrishGillett: > by state do you mean the status field?

I do, sorry for the confusion!

rrxxww: Make two dedicated column for title and status

tap_github/organization_streams.py
```diff
@@ -211,3 +213,820 @@ class TeamRolesStream(GitHubRestStream):
...
+        if context:
+            row["org"] = context["org"]
+        return row
+
+    schema = th.PropertiesList(
+        # Parent keys
+        th.Property("org", th.StringType),
+        # Project fields
+        th.Property(
+            "id", th.StringType, required=False
```
edgarrmondragon, line 300: This is currently functionally equivalent, but it's more semantically adequate IMO. Also, we might decide that `required` doesn't affect whether the field is nullable, only whether it's expected to be present.

```suggestion
            "id", th.StringType, nullable=False
```

edgarrmondragon (commented):
> Current implementation tries to include as many fields as possible available from Graphql.

I think this is best approach.

tap_github/organization_streams.py
rrxxww: Thanks, should I switch to use `nullable` field for all nullable fields then, or just for IDs for now?

rrxxww: Updated the PR to use nullable for id.

edgarrmondragon: Thanks!

edgarrmondragon (approved):

edgarrmondragon:
Thanks @rrxxww and @TrishGillett!

PR #415, opened 2025-06-11
fix: Implement cursor-based pagination and use it for IssuesStream

TrishGillett:
We've begun getting this error for one of our orgs:
> Pagination with the page parameter is not supported for large datasets, please use cursor based pagination (after/before)

I suspect Github has added a new restriction here as we haven't see this issue in the past even with large scale extractions, and it's happening now with a relatively small one.

This PR adds a class variable `use_cursor_pagination` in `GitHubRestStream` that can be toggled to have a given stream subclass use cursor-based pagination instead.  I'm leaving it False by default and only setting it to True in `IssuesStream` because I don't know if other endpoints support it.  If it turns out to be needed in other streams in the future, it should be easy to set.

I've successfully tested this change in my environment.

### DIFF
--- a/tap_github/client.py
+++ b/tap_github/client.py
@@ -42,6 +42,9 @@ class GitHubRestStream(RESTStream):
     # This only has effect on streams whose `replication_key` is `updated_at`.
     use_fake_since_parameter = False
 
+    # Set to True to use cursor-based pagination instead of page-based pagination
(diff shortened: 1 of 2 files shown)


TrishGillett:
Testing complete on my end. ✅

edgarrmondragon (approved):
Thanks!

PR #436, opened 2025-07-24
ADD GraphQL Discussion Streams: Categories, Discussions, Comments, Replies

JeannC:
- Add Github Discussion Streams:
  - `DiscussionCategoriesStream`
  - `DiscussionCommentRepliesStream`
  - `DiscussionCommentsStream`
  - `DiscussionsStream`
- Pass `"has_discussions"` record to children of the Repository Stream
- Create new Schema Object for `reaction_type_object`

Notes:
- `_Disussion Logger` class is temporary for testing only. The class will be deleted pre-ship.
- `self.logger.info` lines will either be switched to debug only, or deleted pre-ship


tap_github/schema_objects.py
```diff
@@ -19,6 +19,18 @@
     th.Property("site_admin", th.BooleanType),
 )
 
+# The actor object is common through GraphQL
+actor_object = th.ObjectType(
```
TrishGillett, line 23: I think in previous graphql prototyping we've re-used the user_object even for graphQL:
- renamed the http_url value to html_url in the graphQL query
- omitted the resource_path from the query
- gravatar_url isn't available from the graphQL API but I think it will just populate nulls in that column if you use user_agent in your schema (needs a sanity check)

tap_github/repository_streams.py
```diff
@@ -1916,6 +1921,808 @@ def query(self) -> str:
...
+    name = "discussion_categories"
+    query_jsonpath = "$.data.repository.discussionCategories.nodes.[*]"
+    primary_keys: ClassVar[list[str]] = [
+        "node_id"
+    ]  # id:databaseId is not available for the categories object
+    replication_key = "updated_at"
+    parent_stream_type = RepositoryStream
+    state_partitioning_keys: ClassVar[list[str]] = ["repo_id"]
+    ignore_parent_replication_key = True  # Repository's updated_at does not change when a new discussion category is added/modified  # noqa: E501
+    use_fake_since_parameter = True
```
TrishGillett, line 1936: Is this intended to extract incrementally or full extract every run?  I think we won't want this setting if it's the latter.

tap_github/repository_streams.py
```diff
@@ -1916,6 +1920,911 @@ def query(self) -> str:
...
+    ignore_parent_replication_key = True  # Repository's updated_at does not change when a new discussion category is added/modified  # noqa: E501
+
+    def get_records(self, context: Context | None = None) -> Iterable[dict[str, Any]]:
+        """
+        Return a generator of row-type dictionary objects.
+        If discussions are not enabled, skip the API call.
+        """
+        repo = context.get("repo", "unknown")
+        org = context.get("org", "unknown")
+        if not context.get("has_discussions", False):
```
TrishGillett, line 1948: Documenting context: This is a nice way to cut down calls in a context where there are lots of repos but only a small fraction are using discussions.

tap_github/repository_streams.py
```diff
@@ -1916,6 +1920,911 @@ def query(self) -> str:
...
+        th.Property("updated_at", th.DateTimeType),
+    ).to_dict()
+
+
+class DiscussionsStream(GitHubGraphqlStream):
+    """
+    Defines stream fetching discussions from each repository.
+
+    Singer SDK does not support smart pagination for GraphQL,
+    so progress for this stream is not resumable if interrupted.
```
TrishGillett, line 2026:
```suggestion
    Much like the `use_fake_since_parameter` setting for REST streams,
    this stream extracts data in descending order and we exit early when
    we've seen records that go back as far as we need.
    
    Note that this stream is not resumable if interrupted.
```

tap_github/repository_streams.py
```diff
@@ -1916,6 +1920,911 @@ def query(self) -> str:
...
+
+class DiscussionCommentRepliesStream(GitHubGraphqlStream):
+    """
+    Defines stream fetching replies for each discussion comment from each repository.
+
+    Edits made to replies after extraction are not reflected in the stream.
+    Full refresh is required to see the changes.
+
+    Singer SDK does not support smart pagination for GraphQL,
+    so progress for this stream is not resumable if interrupted.
```
TrishGillett, line 2571:
```suggestion
    Much like the `use_fake_since_parameter` setting for REST streams,
    this stream extracts data in descending order and we exit early when
    we've seen records that go back as far as we need. Since this object
    doesn't offer an orderBy field, we use tail-first pagination to extract
    in descending order. A side effect of this is that within each page the
    records are still ascending order, so some special handling is needed.
    
    Note that this stream is not resumable if interrupted.
```

PR #455, opened 2025-09-08
fix(client): Extract records via `records_jsonpath`

afkerhousse:
The GitHub API doesn’t consistently return records under an "items" key. Some streams return arrays at the top level, while others (e.g., Workflow Runs) use different shapes/keys (e.g., `workflow_runs`). The previous logic silently returned None for streams without an "items" key, breaking iteration and record emission. This change defers to the stream-defined `records_jsonpath `for non-list responses.

Benefits:
- Respects each stream’s declared records_jsonpath.
- Works for endpoints that don’t use "items" (e.g., WorkflowRuns).
- Keeps existing behavior for endpoints returning a top-level list.
- Removes a brittle, hard-coded assumption about response shape. 

Notes:
- For streams that previously relied on "items", ensure their records_jsonpath targets the same array (e.g., $.items[*]) so behavior remains equivalent.
- Minor performance cost from JSONPath evaluation is outweighed by correctness and consistency.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -154,84 +154,101 @@ files = [
 
 [[package]]
 name = "cffi"
-version = "1.17.1"
+version = "2.0.0"
 description = "Foreign Function Interface for Python calling C code."
 optional = false
-python-versions = ">=3.8"
+python-versions = ">=3.9"
 groups = ["main"]
 markers = "platform_python_implementation != \"PyPy\""
 files = [
-    {file = "cffi-1.17.1-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:df8b1c11f177bc2313ec4b2d46baec87a5f3e71fc8b45dab2ee7cae86d9aba14"},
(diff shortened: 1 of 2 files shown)


edgarrmondragon (commented):
Thanks @afkerhousse!

This seems fairly safe, though I have not double checked every stream but should we also update

https://github.com/MeltanoLabs/tap-github/blob/afbe30f66042d377fed10236ec57b739c6a262b6/tap_github/client.py#L266-L286

?

afkerhousse:
Hi @edgarrmondragon 
Yes I think it needs to be updated as well

`def parse_response(self, response: requests.Response) -> Iterable[dict]:
    """Parse the response and return an iterator of result rows."""
    if response.status_code in (
        [*self.tolerated_http_errors, EMPTY_REPO_ERROR_STATUS]
    ):
        return

    # Update token rate limit info and loop through tokens if needed.
    self.authenticator.update_rate_limit(response.headers)

    resp_json = response.json()
    
    # Handle different response structures
    if isinstance(resp_json, list):
        # Direct array response
        results = resp_json
    else:
        # Use records_jsonpath to extract records
        results = list(extract_jsonpath(self.records_jsonpath, input=resp_json))
        
        # Fallback: if no records found via jsonpath, treat the whole response as a single record
        if not results:
            results = [resp_json]

    yield from results`

PR #555, opened 2026-04-25
fix: Keep org-specific token rotation isolated

stray-nick:
## Summary

Keep token rotation scoped to the current organization when that organization has its own configured token pool.

GitHub App installation tokens are org-scoped. If a stream is reading private repositories for one org and `get_next_auth_token()` falls through to another org's installation token, GitHub can return misleading `404 Not Found` responses for repositories that do exist and are accessible with the correct org token.

## What changed

- `get_next_auth_token()` now prefers the current organization's token managers when `current_organization` has a configured token pool.
- Fallback to org-agnostic or other-org tokens is still allowed when the current organization has no configured token pool.
- Added regression tests for both paths:
  - org-specific token rotation stays inside the current org's pool.
  - missing org-specific pools still use the existing fallback behavior.

## Validation

```text
uv run pytest tests/test_authenticator.py
# 44 passed
```

### DIFF
--- a/tap_github/authenticator.py
+++ b/tap_github/authenticator.py
@@ -525,31 +525,35 @@ class GitHubTokenAuthenticator(APIAuthenticatorBase):
     def get_next_auth_token(self) -> None:
         current_token = self.active_token.token if self.active_token else ""
 
-        # Build a list of candidate tokens for the current organization
+        # Build a list of candidate tokens for the current organization.
+        # If the current org has a configured token pool, stay within that pool.
+        # Installation tokens are org-scoped, so falling through to another org's
+        # token can turn legitimate private-repo calls into misleading 404s.
         candidates = []
-
-        # Priority 1: Other tokens for the current organization
-        if (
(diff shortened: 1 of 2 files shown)


tap_github/authenticator.py
```diff
@@ -525,31 +525,35 @@ def set_organization(self, org: str) -> None:
...
-        # Priority 1: Other tokens for the current organization
-        if (
+        has_current_org_tokens = (
             self.current_organization
             and self.current_organization in self.token_managers
-        ):
+            and self.token_managers[self.current_organization]
+        )
+
+        if has_current_org_tokens:
```
TrishGillett, line 540:
```suggestion
        # Priority 1: Tokens scoped to the current org
        if has_current_org_tokens:
```

TrishGillett (approved):
👌

edgarrmondragon (approved):
Thanks folks!

stray-nick:
Appreciate it, Edgar!

PR #571, opened 2026-06-03
fix: support repositories with pull requests disabled

stray-nick:
## Summary

This PR lets the tap handle repositories that have pull requests disabled.

GitHub can return `"has_pull_requests": false` in a repository payload. For those repositories, `GET /repos/{owner}/{repo}/pulls` can return `404 Not Found` even though the repository itself is readable and other endpoints can still be extracted.

Instead of failing the sync for that case, the tap now skips pull request extraction for repositories where GitHub explicitly reports that pull requests are disabled.

## Behavior

- When `has_pull_requests` is exactly `false`, the tap skips the `/pulls` request for that repository.
- When `has_pull_requests` is `true`, missing, `null`, `0`, or otherwise unknown, behavior is unchanged: the tap still requests `/pulls` and surfaces GitHub errors normally.
- The change applies only to pull request extraction.

## Implementation

- Includes GitHub's `has_pull_requests` field in repository records.
- Carries that flag to the pull request extraction step.
- Checks the flag before requesting `/repos/{owner}/{repo}/pulls`.

## What this does not change

- It does not add broad `404` or `403` tolerance.
- It does not change issue extraction or infer pull request availability from `has_issues`.
- It does not change discussion extraction.

## Validation

```text
LOGLEVEL=WARNING uv run pytest tests/test_tap.py -k 'pull_request_capability or pull_requests_stream or issues_stream'
# 9 passed, 8 deselected

uv run ruff check tap_github/repository_streams.py tests/test_tap.py
# All checks passed

uv run mypy tap_github
# Success: no issues found in 12 source files

uv run ty check tap_github
# All checks passed

git diff --check
# passed
```

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -335,6 +335,40 @@ class RepositoryStream(GitHubRestStream):
     ).to_dict()
 
 
+class RepositoryFeatureGatedStream(GitHubRestStream):
+    """REST stream that skips API calls for explicitly disabled repo features."""
+
+    required_repository_features: ClassVar[tuple[str, ...]] = ()
+
+    def _disabled_repository_feature(self, context: Context | None) -> str | None:
+        """Return the first explicitly disabled repository feature, if any."""
+        if context is None:
+            return None
+
+        for feature in self.required_repository_features:
+            if context.get(feature) is False:
+                return feature
+
+        return None
+
+    def get_records(self, context: Context | None = None) -> Iterable[dict[str, Any]]:
+        """Return records unless repository metadata explicitly disables a feature."""
+        if disabled_feature := self._disabled_repository_feature(context):
+            org = context.get("org", "unknown") if context is not None else "unknown"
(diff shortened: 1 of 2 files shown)


tap_github/repository_streams.py
```diff
@@ -1302,6 +1304,27 @@ class PullRequestsStream(GitHubRestStream):
...
+        repo = context.get("repo", "unknown")
+        org = context.get("org", "unknown")
+        if context.get("has_pull_requests", True) is False:
+            self.logger.debug(
+                f"Repository {org}/{repo}: Pull requests not enabled, "
+                "skipping API call",
+            )
+            return []
+
+        self.logger.debug(
```
TrishGillett: My two cents: logs already tend to get very long, IMO we could keep the logging above for the case where it's not enabled, but remove this logging for the 'normal' case.

stray-nick: That's a good call. I'll make that change.

edgarrmondragon (approved):
LGTM

PR #572, opened 2026-06-03
feat: Add configurable page size for `project_items` stream

KennedyMurphy:
## Summary

Add a `stream_options.project_items.page_size` config option for the `project_items` stream.

The default remains `100`, matching the current hardcoded GitHub GraphQL page size and preserving existing behavior. Users can lower the value, for example to `50`, to reduce per-request GraphQL query complexity for large Projects-v2 boards with many custom fields.

Example:

```json
{
  "stream_options": {
    "project_items": {
      "page_size": 50
    }
  }
}
```

## Motivation

`ProjectItemsStream` dynamically includes all known project fields in each item query. For large GitHub Projects-v2 boards with many custom fields, `items(first: 100, ...)` can produce expensive GraphQL requests that are more likely to hit GitHub timeout/query-complexity behavior.

Lowering the page size keeps cursor pagination semantics unchanged, so the result set is preserved, while reducing the amount of data and field-value expansion requested per GraphQL call. The tradeoff is more requests.

## Validation

- `uv run pytest tests/test_tap.py -q -k project_items`
- `uvx ruff check tap_github tests/test_tap.py`
- `uv run mypy tap_github`

Note: `uv run ty check tap_github` was attempted locally but the process was killed with exit code 137.

---
_Generated with Pi._


edgarrmondragon (approved):
Thanks @KennedyMurphy!

PR #582, opened 2026-06-15
fix: skip unavailable issue comments endpoint

johncomar:
## Summary

This PR lets the tap handle repositories where GitHub disables the repository-level issue comments endpoint.

GitHub can return `"has_issues": false` and `"has_pull_requests": false` in a repository payload. For those repositories, `GET /repos/{owner}/{repo}/issues/comments` can return `404 Not Found` even though the repository itself is readable and other endpoints can still be extracted.

Instead of failing the sync for that case, the tap now skips issue comment extraction for repositories where GitHub explicitly reports that both Issues and Pull Requests are disabled.

This follows the same conservative pattern added for repositories with pull requests disabled in #571.

## Behavior

- When `has_issues` and `has_pull_requests` are both exactly `false`, the tap skips the `/issues/comments` request for that repository.
- When either flag is `true`, missing, `null`, `0`, or otherwise unknown, behavior is unchanged: the tap still requests `/issues/comments` and surfaces GitHub errors normally.
- The change applies only to issue comment extraction.

## Implementation

- Includes GitHub's `has_issues` field in repository records.
- Carries that flag to repository child stream context alongside `has_pull_requests`.
- Checks both flags before requesting `/repos/{owner}/{repo}/issues/comments`.

## Related stream audit

This PR does not broadly gate every issue-adjacent stream.

The issue comments endpoint can include pull request conversation comments because GitHub models pull requests as issues. For that reason, the tap skips only when GitHub explicitly reports that both Issues and Pull Requests are disabled.

The following streams were intentionally left unchanged and continue to delegate normally when `has_issues` is `false`:

- `issues`
- `issue_events`
- `milestones`
- `labels`

The focused change is limited to `issue_comments`, where the repository-level issue comments endpoint can fail when both Issues and Pull Requests are disabled.

## What this does not change

- It does not add broad `404` or `403` tolerance.
- It does not suppress generic GitHub errors.
- It does not change pull request extraction.
- It does not change discussion extraction.

## Validation

```text
LOGLEVEL=WARNING uv run pytest tests/test_tap.py -k 'has_issues or issue_comments or issue_capability'
# 16 passed, 18 deselected

uv run --with ruff==0.15.15 ruff check tap_github/repository_streams.py tests/test_tap.py
# All checks passed

uv run mypy tap_github
# Success: no issues found in 12 source files

uv run ty check tap_github
# All checks passed

git diff --check
# passed
```

### DIFF
--- a/tap_github/repository_streams.py
+++ b/tap_github/repository_streams.py
@@ -243,6 +243,7 @@ class RepositoryStream(GitHubRestStream):
             "has_discussions": record.get(
                 "has_discussions", False
             ),  # GitHub repos not updated after the feature was released in 2021 will not have this field. # noqa: E501
+            "has_issues": record.get("has_issues", True),
             "has_pull_requests": record.get("has_pull_requests", True),
         }
 
@@ -325,6 +326,7 @@ class RepositoryStream(GitHubRestStream):
         th.Property("network_count", th.IntegerType),
         th.Property("subscribers_count", th.IntegerType),
         th.Property("open_issues_count", th.IntegerType),
+        th.Property("has_issues", th.BooleanType),
         th.Property("has_pull_requests", th.BooleanType),
         th.Property("allow_squash_merge", th.BooleanType),
         th.Property("allow_merge_commit", th.BooleanType),
@@ -1018,6 +1020,19 @@ class IssueCommentsStream(GitHubRestStream):
 
         Each row emitted should be a dictionary of property names to their values.
         """
+        if (
+            context
+            and context.get("has_issues", True) is False
+            and context.get("has_pull_requests", True) is False
+        ):
+            repo = context.get("repo", "unknown")
+            org = context.get("org", "unknown")
+            self.logger.debug(
+                f"Repository {org}/{repo}: Issues and pull requests not enabled, "
+                "skipping issue comments API call",
+            )
+            return []
+
         if context and context.get("comments", None) == 0:
             self.logger.debug(f"No comments detected. Skipping '{self.name}' sync.")
             return []
--- a/tests/test_tap.py
+++ b/tests/test_tap.py
@@ -189,6 +189,43 @@ def test_repository_child_context_defaults_pull_request_capability_to_enabled(
     assert context["has_pull_requests"] is True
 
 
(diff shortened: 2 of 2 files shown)


tap_github/repository_streams.py
```diff
@@ -1018,6 +1020,19 @@ def get_records(self, context: Context | None = None) -> Iterable[dict[str, Any]
 
         Each row emitted should be a dictionary of property names to their values.
         """
+        if (
+            context
+            and context.get("has_issues", True) is False
+            and context.get("has_pull_requests", True) is False
```
TrishGillett, line 1026: Good catch to recognize this stream is valid if either is True!

stray-nick (approved):
Looks like this follows https://github.com/MeltanoLabs/tap-github/pull/571 exactly, but for issues.

Only comment I have is given this is a second occurrence of the same issue, should we extend it to other "potentially-disableable" github artifacts now? cc @TrishGillett

tap_github/repository_streams.py
edgarrmondragon: Is that because `issue_comments` also pull PR comments?

johncomar: Yes, that's correct. [The GitHub docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10) show how pull requests are treated as a type of issue.

This stream reads the repo-wide issue comments API, so it can include normal PR conversation comments as well as comments on standalone issues. That’s why this change only skips when both Issues and Pull Requests are turned off.

edgarrmondragon: Gotcha, thanks for confirming!

edgarrmondragon (approved):

tap_github/repository_streams.py
TrishGillett: Yes!  For a pull request, comments written using the comment box at the bottom of the `Conversation` page (front page) of the PR are considered [issue comments](https://docs.github.com/en/rest/issues/comments?apiVersion=2026-03-10), and comments attached to the diff on the `Files changed` page (or their replies) are considered [review comments](https://docs.github.com/en/rest/pulls/comments?apiVersion=2026-03-10).

